Issue #2, opened 2021-07-24
Allow shell glob patterns for label/annotation keys


ymmt2005:
## What

Currently, Accurate only propagates labels and annotations whose key matches
exactly with one of the keys defined in the configuration file.

It'd be handy if it allows shell glob patterns such as `*.cybozu.com/*`.

## How

Use [path.Match](https://pkg.go.dev/path#Match).

## Checklist

- [x] Finish implementation of the issue
- [x] Test all functions
- [x] Have enough logs to trace activities
- [x] Notify developers of necessary actions

PR #8, opened 2021-08-03
Allow shell glob patters for label/annotation keys.

d-kuro:
ref: https://github.com/cybozu-go/accurate/issues/2

### DIFF
--- a/controllers/namespace_controller.go
+++ b/controllers/namespace_controller.go
@@ -3,6 +3,7 @@ package controllers
 import (
 	"context"
 	"fmt"
+	"path/filepath"
 	"reflect"
 
 	"github.com/cybozu-go/accurate/pkg/constants"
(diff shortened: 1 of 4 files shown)


controllers/namespace_controller.go
```diff
@@ -76,16 +77,30 @@ func (r *NamespaceReconciler) reconcile(ctx context.Context, ns *corev1.Namespac
 func (r *NamespaceReconciler) propagateMeta(ctx context.Context, ns, parent *corev1.Namespace) error {
 	orig := ns.DeepCopy()
 	for _, key := range r.LabelKeys {
```
ymmt2005, line 79: Would you refactor the logic so that it would look like:

```go
for k, v := range parent.Labels {
    if r.matchLabelKey(k) {
        ns.Labels[k] = v
    }
}
```

controllers/namespace_controller.go
```diff
@@ -76,16 +77,30 @@ func (r *NamespaceReconciler) reconcile(ctx context.Context, ns *corev1.Namespac
...
-		if val, ok := parent.Labels[key]; ok {
-			ns.Labels[key] = val
+		for label, val := range parent.Labels {
+			ok, err := filepath.Match(key, label)
+			if err != nil {
```
ymmt2005, line 82: The only possible error here is `filepath.ErrBadPattern` here,
and such a bad pattern can be eliminated beforehand when
loaded from the configuration file.

Please validate the pattern when loaded.

controllers/namespace_controller.go
```diff
@@ -76,16 +77,30 @@ func (r *NamespaceReconciler) reconcile(ctx context.Context, ns *corev1.Namespac
...
+			if ok {
+				ns.Labels[label] = val
+			}
 		}
 	}
 	for _, key := range r.AnnotationKeys {
-		if val, ok := parent.Annotations[key]; ok {
-			if ns.Annotations == nil {
-				ns.Annotations = make(map[string]string)
+		for annotation, val := range parent.Annotations {
```
ymmt2005, line 92: Likewise.

ymmt2005 (changes requested):
Thanks for the fix.
Please address my comments, and update the user manual too.

controllers/namespace_controller.go
d-kuro: Thank you comments.
I fixed:
https://github.com/cybozu-go/accurate/pull/8/commits/88a1a9a6e7dbf4dbf805057a465ea56732d6942e

controllers/namespace_controller.go
d-kuro: Thanks.
fixed:
https://github.com/cybozu-go/accurate/pull/8/commits/dd7cf644f4fa7fa5a7d1c164888a97130fb3963a

controllers/namespace_controller.go
d-kuro: fixed:
dd7cf64

ymmt2005 (changes requested):
I'm sorry, but we should use `path.Match` instead of `filepath.Match` for labels and annotations.
https://pkg.go.dev/path#Match

Please change the package, and also update the user manual.

d-kuro:
@ymmt2005 Thank you follow up. I fixed:
https://github.com/cybozu-go/accurate/pull/8/commits/d66c3316f63cc742d1585fab1d1e28df7649f474

ymmt2005 (approved):
LGTM

Issue #7, opened 2021-07-29
Set labels/annotations of sub-namespaces


ymmt2005:
## What

Currently, Accurate does not allow tenant users to set new labels/annotations to sub-namespaces created from SubNamespace objects.

It is handy to allow tenant users to set new labels/annotations only when:

1. the label/annotation key is included in the configuration file, and
2. the label/annotation is not set for the parent namespace.

## How

1. Add new fields to SubNamespace, for example, `spec.labels` and `spec.annotations`
2. Enhance `accurate-controller` to create a sub-namespace with the specified labels/annotations.

`accurate-controller` has to ignore labels/annotations whose key is not listed in the configuration file.

It also should not set labels/annotations if the same labels/annotations are set for the parent namespace.
Nevertheless, the value would be overwritten by the existing Namespace controller to the value of the parent labels/annotations.

Describe how to address the issue.

## Checklist

- [x] Finish implementation of the issue
- [x] Test all functions
- [x] Have enough logs to trace activities
- [x] Notify developers of necessary actions

ymmt2005:
implemented in #19.

PR #19, opened 2021-09-23
Add labels and annotations field to SubNamespace resource

bells17:
#7


api/v1/subnamespace_types.go
```diff
@@ -16,13 +16,22 @@ const (
 	SubNamespaceConflict = SubNamespaceStatus("conflict")
 )
 
+type SubNamespaceSpec struct {
```
ymmt2005, line 19: Please add comments to the struct and every field for documentation purposes.

api/v1/subnamespace_types.go
```diff
@@ -16,13 +16,22 @@ const (
...
+	Labels      map[string]string `json:"labels,omitempty"`
+	Annotations map[string]string `json:"annotations,omitempty"`
```
ymmt2005, line 21: These should have `// +optional` comments.

controllers/subnamespace_controller.go
```diff
@@ -149,3 +162,16 @@ func (r *SubNamespaceReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		}).
 		Complete(r)
 }
+
+func cloneMap(m map[string]string, excludeKeys []string) map[string]string {
```
ymmt2005, line 166: This does not implement the proposed specification.

> accurate-controller has to ignore labels/annotations whose key is not listed in the configuration file.

ymmt2005 (changes requested):

zoetrope (commented):
I think this feature should be used with caution, because you may get privileges by setting labels or annotations.
For example, [PodSecurity Admission](https://github.com/kubernetes/enhancements/tree/master/keps/sig-auth/2579-psp-replacement) will control policies using namespace labels.

So, please write a note in the documentation like the following:
If you are using something that controls permissions by labels or annotaions, such as PodSecurity Admission, an administrator should set root namespaces appropriate labels and annotations.

ymmt2005:
@zoetrope 
Accurate only sets the labels and annotations that are explicitly permitted
in the configuration file.   So, admins have full control over what a tenant
user can do.

ymmt2005:
@bells17 @zoetrope 
One thing that I haven't determined is that whether we should have a different set of
label/annotation keys for this feature than the keys for propagation.

What do you think?

zoetrope:
I understand.

I prefer a simple configuration.
I think we should not have the different set.

bells17:
By the way, should we prioritize SubNamespace labels/annotations over parental Namespace labels/annotations if there have the same labels/annotations key?

ymmt2005:
OK, let's go with the current proposal.

ymmt2005:
@bells17 
> By the way, should we prioritize SubNamespace labels/annotations over parental Namespace labels/annotations if there have the same labels/annotations key?

Definitely not.  And even if they would conflict, the parent Namespace labels/annotations win because they are propagated later on.

bells17:
Thank you @ymmt2005 @zoetrope .
I've corrected the points you commented on.
Would you review this PR again?

charts/accurate/MIGRATION.md
```diff
@@ -70,6 +70,18 @@ controller:
     # https://metallb.universe.tf/usage/#requesting-specific-ips
     # - metallb.universe.tf/address-pool
 
+    # controller.config.subNamespaceLabelKeys -- Labels to be propagated to sub-namespaces from SubNamespace resource.
+    # It is also possible to specify a glob pattern that can be interpreted by Go's "path.Match" func.
+    # https://pkg.go.dev/path#Match
+    subNamespaceLabelKeys: []
+    # - app
+
+    # controller.config.subNamespaceAnnotationKeys -- Labels to be propagated to sub-namespaces from SubNamespace resource.
```
zoetrope, line 79:
```suggestion
    # controller.config.subNamespaceAnnotationKeys -- Annotations to be propagated to sub-namespaces from SubNamespace resource.
```

charts/accurate/values.yaml
```diff
@@ -40,6 +40,18 @@ controller:
     # https://metallb.universe.tf/usage/#requesting-specific-ips
     # - metallb.universe.tf/address-pool
 
+    # controller.config.subNamespaceLabelKeys -- Labels to be propagated to sub-namespaces from SubNamespace resource.
+    # It is also possible to specify a glob pattern that can be interpreted by Go's "path.Match" func.
+    # https://pkg.go.dev/path#Match
+    subNamespaceLabelKeys: []
+    # - app
+
+    # controller.config.subNamespaceAnnotationKeys -- Labels to be propagated to sub-namespaces from SubNamespace resource.
```
zoetrope, line 49:
```suggestion
    # controller.config.subNamespaceAnnotationKeys -- Annotations to be propagated to sub-namespaces from SubNamespace resource.
```

docs/crd_subnamespace.md
```diff
@@ -28,3 +30,14 @@ SubNamespaceList contains a list of SubNamespace
...
 [Back to Custom Resources](#custom-resources)
+
+#### SubNamespaceSpec
+
+SubNamespaceSpec defines the desired state of SubNamespace
+
+| Field | Description | Scheme | Required |
+| ----- | ----------- | ------ | -------- |
+| labels | Labels are the labels to be propagated to the sub-namespace | map[string]string | false |
+| annotations | Annotations are the annotations to be propagated to the sub-namespace. | map[string]string | false |
```
zoetrope, line 41: I don't think `propagate` is an appropriate description.
Because it just sets the labels/annotations of SubNamespace to the target Namespace.
Those labels/annotations are not propagated to the descendant namespaces.

ymmt2005: @zoetrope point is valid.
But it would be better to change the implementation to actually propagate
the labels/annotations from SubNamespace to the target Namespace so that
a tenant user can edit Namespace labels/annotations by editing SubNamespaces.

What do you think? @zoetrope @bells17

zoetrope: @ymmt2005 
Perhaps my point was misleading.
The current implementation propagates the labels/annotations from SubNamespace to the target Namespaces.

zoetrope:
@bells17 
I'm curious what happens when I set the same key for labelKeys and subNamespaceLabelKeys.

docs/crd_subnamespace.md
ymmt2005: @zoetrope 
> The current implementation propagates the labels/annotations from SubNamespace to the target Namespaces.

I don't think so.

Take a look at this:
https://github.com/cybozu-go/accurate/blob/add-labels-and-annotations-field2/controllers/subnamespace_controller.go#L95

The SubNamespace controller only creates the Namespace and never updates it afterwards.

zoetrope: @ymmt2005 
Sorry, I understood.

@bells17 
Please fix it to update the labels and annotations in reconcile.

bells17:
@zoetrope 
> I'm curious what happens when I set the same key for labelKeys and subNamespaceLabelKeys.

I implemented so that SubNamespace `spec.labels/spec.annotations` values are preferred more than parent namespace `labels/annotations` values because I think a case of using SubNamespace `spec.labels/spec.annotations` instead of parent namespace `labels/annotations` is a rare case.

bells17:
@ymmt2005 @zoetrope The following points have been updated:

- SubNamespace `spec.labels/spec.annotations` is now propagated to descendants namespaces.
- Added `--labels` and `--annotations` options to the `kubectl-accurate sub create` command.
- Wrote docs.

Would you review these points?

controllers/namespace_controller.go
```diff
@@ -351,7 +374,38 @@ func (r *NamespaceReconciler) reconcileTemplateNamespace(ctx context.Context, ns
...
+				subNSHandler(ev.Object, q)
+			},
+			UpdateFunc: func(ev event.UpdateEvent, q workqueue.RateLimitingInterface) {
+				if ev.ObjectNew.GetDeletionTimestamp() != nil {
+					return
+				}
+				subNSHandler(ev.ObjectOld, q)
+			},
+			DeleteFunc: func(ev event.DeleteEvent, q workqueue.RateLimitingInterface) {
+				subNSHandler(ev.Object, q)
```
zoetrope, line 396: This is not necessary, because when SubNamespace is deleted, the target Namespace will be also deleted.

zoetrope (commented):
I read the code.
I think the following spec is fine.

> - SubNamespace spec.labels/spec.annotations is now propagated to descendants namespaces.
> - SubNamespace spec.labels/spec.annotations values are preferred more than parent namespace labels/annotations values.

zoetrope (dismissed):
LGTM

cmd/kubectl-accurate/sub/sub_create.go
```diff
@@ -36,6 +38,8 @@ This effectively creates a namespace named NAME as a sub-namespace of NS.`,
 		},
 	}
 
+	cmd.Flags().StringToStringVar(&opts.labels, "labels", opts.labels, "the labels to be propagated to the sub-namespace")
```
ymmt2005, line 41: Please add an example of the usage to the description.  e.g.,
```suggestion
	cmd.Flags().StringToStringVar(&opts.labels, "labels", opts.labels, "the labels to be propagated to the sub-namespace.  Example: a=b,c=d")
```

cmd/kubectl-accurate/sub/sub_create.go
```diff
@@ -36,6 +38,8 @@ This effectively creates a namespace named NAME as a sub-namespace of NS.`,
...
+	cmd.Flags().StringToStringVar(&opts.annotations, "annotations", opts.annotations, "the annotations to be propagated to the sub-namespace")
```
ymmt2005, line 42: Likewise

docs/config.md
```diff
@@ -53,7 +65,8 @@ watches:
 namingPolicies: []
 ```
 
-Only labels and annotations specified in the configuration file will be inherited.
+Only labels and annotations specified in the configuration file will be inherited.  
+If you are using something that controls permissions by labels or annotations, such as PodSecurity Admission, an administrator should allow appropriate labels and annotations.
```
ymmt2005, line 69: I don't see what this means.  Do you mean something like the following?

```suggestion
Be careful that some labels or annotations affect security configurations or the system.
For example, [`pod-security.kubernetes.io/*`](https://kubernetes.io/docs/concepts/security/pod-security-admission/#pod-security-admission-labels-for-namespaces) labels control the security capabilities of Pods in a Namespace.
```

ymmt2005 (changes requested):
Please validate the label keys, label values, and annotation keys in SubNamespace spec field.
If a SubNamespace is created with an invalid key or value, the current implementation would fall into an infinite error loop.

Use [validation.IsDNS1123Label](https://pkg.go.dev/k8s.io/apimachinery/pkg/util/validation#IsDNS1123Label) to validate keys and [validation.IsValidLabelValue](https://pkg.go.dev/k8s.io/apimachinery/pkg/util/validation#IsValidLabelValue) to validate label values.

Also, could you update this part of the user manual for this new feature?
https://cybozu-go.github.io/accurate/subnamespaces.html#creating-a-sub-namespace

controllers/namespace_controller.go
```diff
@@ -92,6 +100,27 @@ func (r *NamespaceReconciler) propagateMeta(ctx context.Context, ns, parent *cor
 			ns.Annotations[k] = v
 		}
 	}
+
+	subNS := &accuratev1.SubNamespace{}
+	err := r.Get(ctx, types.NamespacedName{Name: ns.Name, Namespace: parent.Name}, subNS)
```
ymmt2005, line 105: At the [line 347](https://github.com/cybozu-go/accurate/blob/1302f9f8eb9414b408fb34f466223b8649b94420/controllers/namespace_controller.go#L347), this method is called for a template instance namespace, not a sub-namespace.

I think this should look up a SubNamespace only when `ns` is a sub-namespace (i.e., it has `accurate.cybozu.com/parent` label).

bells17:
> Also, could you update this part of the user manual for this new feature?
https://cybozu-go.github.io/accurate/subnamespaces.html#creating-a-sub-namespace

I've already updated the user manual, but should I need to make additional updates?
https://github.com/cybozu-go/accurate/pull/19/files#diff-b394f5d0e85c3cd9c802a88109196f87c4f5047d04b50c0c557476474460adc4R88-R118

ymmt2005:
@bells17 
> I've already updated the user manual, but should I need to make additional updates?

My bad.  I overlooked that.  It's enough, thank you.

controllers/namespace_controller.go
bells17: fixed.

docs/config.md
bells17: applied.

cmd/kubectl-accurate/sub/sub_create.go
bells17: added.

cmd/kubectl-accurate/sub/sub_create.go
bells17: added.

bells17:
> Please validate the labels and annotations in SubNamespace spec field.

added.

controllers/namespace_controller.go
```diff
@@ -101,11 +101,13 @@ func (r *NamespaceReconciler) propagateMeta(ctx context.Context, ns, parent *cor
...
-	err := r.Get(ctx, types.NamespacedName{Name: ns.Name, Namespace: parent.Name}, subNS)
-	if err != nil && !apierrors.IsNotFound(err) {
-		return fmt.Errorf("failed to get sub namespace %s/%s: %w", ns.Name, parent.Name, err)
-	} else {
+	if _, ok := ns.Labels[constants.LabelParent]; ok {
+		subNS := &accuratev1.SubNamespace{}
+		err := r.Get(ctx, types.NamespacedName{Name: ns.Name, Namespace: parent.Name}, subNS)
+		if err != nil && !apierrors.IsNotFound(err) {
+			return fmt.Errorf("failed to get sub namespace %s/%s: %w", ns.Name, parent.Name, err)
+		}
```
ymmt2005, line 109: What happens if `apierrors.IsNotFound(err)` is `true`?
In that case, subNS is invalid.  Although the following code would work for such invalid data,
I want to check them and skip code using `subNS`.

ymmt2005 (changes requested):

ymmt2005 (approved):
LGTM

PR #35, opened 2021-11-25
Propagate ServiceAccount w/o secrets field

### LINKED ISSUE #33: Propagating ServiceAccount
## What

The current implementation can propagate ServiceAccounts, but the propagated ServiceAccounts
won't work because the Secret tokens issued for them in the parent namespace are not propagated.

The issued Secrets do not have `metadata.ownerRerefences`, so `accurate.cybozu.com/propagate-generated` cannot be used either.

## How

Propagate ServiceAccounts without `secrets` field.

## Checklist

- [x] Finish implementation of the issue
- [x] Test all functions
- [x] Have enough logs to trace activities
- [x] Notify developers of necessary actions

ymmt2005:
to fix #33.


zoetrope (approved):
LGTM

PR #39, opened 2021-12-06
Bump version to 0.3.0

zoetrope:
Signed-off-by: zoetrope <a.ikezoe@gmail.com>


kustomization.yaml
```diff
@@ -10,4 +10,4 @@ configMapGenerator:
 
 images:
 - name: ghcr.io/cybozu-go/accurate
-  newTag: 0.2.0
+  newTag: 0.3.0
```
ymmt2005, line 13: Is this file still referenced in the user manual?
If not, why don't we remove this?

zoetrope: I cut corners.
I must release accurate and helm respectively.

ymmt2005: @zoetrope 
My point is not about Helm.  Can we remove this `kustomizatoin.yaml`?

ymmt2005: In fact, this is not working already because `config.yaml` is missing.

```yaml
resources:
- config/default

# Edit config.yaml to customize Accurate configurations
configMapGenerator:
- name: accurate-config
  namespace: accurate
  files:
  - config.yaml
```

zoetrope: I was mistaken.
`kustomization.yaml` is not used and will be removed.

ymmt2005 (approved):
LGTM

PR #43, opened 2022-01-25
Fix the broken chart

zoetrope:
kustomize 4.0.5 or later has a issue with long lines being broken.
https://github.com/kubernetes-sigs/kustomize/issues/3969

This will cause `helm install` to fail.

Signed-off-by: zoetrope <a.ikezoe@gmail.com>


ymmt2005 (approved):
LGTM

PR #47, opened 2022-02-03
docs: clarify the behavior of label/annotation propagation

Hsn723:
The manual states that Accurate propagates labels/annotations, however (although it makes sense) it is not plainly obvious that Accurate only propagates labels/annotations that have been configured in that matter via config.yaml.

This PR adds some minor clarifications in the manual to prevent such misunderstandings


ymmt2005 (approved):
LGTM.  Thank you!

PR #49, opened 2022-02-03
prevent accurate's label/annotation prefix from being selected for propagation

### LINKED ISSUE #48: Accurate can be erroneously set-up to propagate its own labels/annotations
**Describe the bug**
Admitedly, I did not yet try to reproduce the issue as I am not sure if this is intended behavior or not, but currently when configuring Accurate's `labelKeys`/`annotationKeys`, no check is performed as to whether Accurate's own [labels](https://cybozu-go.github.io/accurate/labels.html) and [annotations](https://cybozu-go.github.io/accurate/annotations.html) have been specified.

For instance, `config.Validate` only checks if the labels/annotations are syntatically valid:
https://github.com/cybozu-go/accurate/blob/6773c56d67c2a180c93aadb30b096d4d9e587232/pkg/config/types.go#L37-L43

As a result, an administrator could set up Accurate's config.yaml to propagate Accurate's own labels and annotations and it would happily oblige.

Accurate controller's `propagateMeta` will indeed propagate all configured labels/annotations:
https://github.com/cybozu-go/accurate/blob/6773c56d67c2a180c93aadb30b096d4d9e587232/controllers/namespace_controller.go#L90-L94

Granted, this is purely user misconfiguration, though I think that adding a check for the mistaken specification of labels/annotations in the [`constants.MetaPrefix`](https://github.com/cybozu-go/accurate/blob/a8bdffc1fd52e06bc8073bdc7686f277bb33729c/pkg/constants/meta.go#L4) namespace (or specific labels/annotations that should definitely not be configured to propagate?) would be more user-friendly.

**Environments**
- Version:
- OS:

**To Reproduce**
Steps that should reproduce the behavior:
1. Add the following to `config.yaml`
```yaml
labelKeys:
- accurate.cybozu.com/type
```
2. Deploy Accurate
3. Add a SubNamespace
4. It should have inherited the `accurate.cybozu.com/type=root` from its parent namespace

**Expected behavior**
I am not exactly sure if this is expected behavior, but it does not seem like it is originally intended to happen.

**Additional context**

Hsn723:
Closes #48

### DIFF
--- a/pkg/config/types.go
+++ b/pkg/config/types.go
@@ -4,11 +4,14 @@ import (
 	"fmt"
 	"path"
 	"regexp"
+	"strings"
 
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"sigs.k8s.io/yaml"
+
+	"github.com/cybozu-go/accurate/pkg/constants"
 )
 
 // NamingPolicy represents naming policies for Namespaces created from SubNamespaces.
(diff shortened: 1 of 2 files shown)


ymmt2005 (approved):
LGTM

PR #51, opened 2022-02-03
clarify label/annotation deletion behavior in docs and add test case

### LINKED ISSUE #50: Label/annotation deletion is not propagated to child namespaces
**Describe the bug**
When a propagated label/annotation is deleted from a parent namespace, the deletion is not propagated to its child namespaces.

**Environments**
- Version:
- OS:

**To Reproduce**
1. Configure Accurate to propagate the `baz.glob/c` label via config.yaml
2. Deploy Accurate
3. Create a root namespace with the `baz.glob/c` label and a child namespace
4. Delete the `baz.glob/c` label from the root namespace
5. Observe that the label is still present in the child namespace

**Expected behavior**
From a developer somewhat familiar with the inner workings of the controller's perspective, it makes sense that propagating deletions of labels/annotations is non-trivial. Furthermore, the absence of a label/annotation on the parent (via deletion of omission) does not necessarily imply that it must be deleted on the child namespace, so user intent is also hard to grasp.

On the other hand, from a reader's perspective, `propagation` means to me that creation, update and deletion is reconciled.

This can be addressed either by:

- implementing the propagation of label/annotation deletion
- updating the manual to clarify that the propagation of label/annotation deletions is not supported

As a predictable implementation of deletion propagation is non-trivial and may break user expectations, I am leaning towards the latter. Any opinions?

**Additional context**

Hsn723:
Closes #50

### DIFF
--- a/controllers/namespace_controller_test.go
+++ b/controllers/namespace_controller_test.go
@@ -376,6 +376,7 @@ var _ = Describe("Namespace controller", func() {
 		root.Annotations = map[string]string{
 			"foo":                      "bar",
 			"bar.glob/b":               "glob",
+			"baz.glob/c":               "delete-me",
 			"do.not.match/glob.patten": "glob",
 		}
 		err := k8sClient.Create(ctx, root)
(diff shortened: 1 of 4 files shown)


ymmt2005 (approved):
LGTM.  Thank you!

PR #54, opened 2022-07-05
e2e: fix instability

ymmt2005:
sealed-secret-controller can take some time to create a secret
necessary for further operations.


e2e/Makefile
```diff
@@ -35,6 +35,9 @@ start: $(KIND) $(KUBECTL) $(HELM) $(KUBESEAL) $(KUBECTL_ACCURATE)
 	$(HELM) repo update
 	$(HELM) install sealed-secrets-controller --namespace kube-system sealed-secrets/sealed-secrets
 	$(KUBECTL) -n kube-system wait --for=condition=available --timeout=180s --all deployments
+	$(KUBECTL) -n kube-system wait --for=condition=available --timeout=180s --all deployments
```
zoetrope, line 38: Do you need the same command twice?

ymmt2005: No. That was my mistake. will fix.

zoetrope (approved):
LGTM

Issue #55, opened 2022-07-07
Support Kubernetes 1.24


masa213f:
## What

Support Kubernetes 1.24

## How

Describe how to address the issue.

## Checklist

- [ ] Finish implementation of the issue
- [ ] Test all functions
- [ ] Have enough logs to trace activities
- [ ] Notify developers of necessary actions

PR #57, opened 2022-07-21
Support Kubernetes 1.24

masa213f:
https://github.com/cybozu-go/accurate/issues/55
Support Kubernetes 1.24 and update dependencies.

- Update go modules.
    - Update ginkgo to v2
- Update build and test tools.
    - Stop using nilerr.
- Update actions.
- Build with go 1.18

And I update e2e test due to the following change in k8s 1.24.

https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.24.md#urgent-upgrade-notes
>The LegacyServiceAccountTokenNoAutoGeneration feature gate is beta, and enabled by default. When enabled, Secret API objects containing service account tokens are no longer auto-generated for every ServiceAccount. Use the [TokenRequest](https://kubernetes.io/docs/reference/kubernetes-api/authentication-resources/token-request-v1/) API to acquire service account tokens, or if a non-expiring token is required, create a Secret API object for the token controller to populate with a service account token by following this [guide](https://kubernetes.io/docs/concepts/configuration/secret/#service-account-token-secrets).

Signed-off-by: Masayuki Ishii <masa213f@gmail.com>


Makefile
```diff
@@ -74,9 +73,9 @@ check-generate:
 .PHONY: envtest
 envtest: setup-envtest
 	source <($(SETUP_ENVTEST) use -p env); \
-		TEST_CONFIG=1 go test -v -count 1 -race ./pkg/config -ginkgo.progress -ginkgo.v -ginkgo.failFast
+		TEST_CONFIG=1 go test -v -count 1 -race ./pkg/config -ginkgo.progress -ginkgo.v -ginkgo.fail-fast
```
masa213f, line 76: --ginkgo.failFast is deprecated, use --ginkgo.fail-fast instead
Learn more at:https://onsi.github.io/ginkgo/MIGRATING_TO_V2#changed-command-line-flags

e2e/e2e_test.go
```diff
@@ -307,6 +335,32 @@ var _ = Describe("kubectl accurate", func() {
...
+		if k8sMinorVersion < 24 {
+			Skip("this test case does not work")
+		}
+
+		// From Kubernetes 1.24, the auto-generation of secret-based service account tokens has been disabled by default.
+		// So the secrets field in the ServiceAccount is not updated. But when upgrading Kubernetes from 1.23 or lower,
+		// some ServiceAccounts that have been created before the upgrade might have the secrets field.
+		// In this case, accurate should not copy the field.
+		kubectlSafe(serviceAccountWithDummySecretsYAML, "apply", "-f", "-")
+		Consistently(func() error {
```
morimoto-cybozu, line 348: The creation of `sn1/test` should be waited with `Eventually()`.
This `Consistently()` can fail if the creation of `sn1/test` ServiceAccount is delayed.
As far as I know `kubectl apply` above is synchronous only for `subroot1/test`, not for `sn1/test`.

Once `sn1/test` is created, it would be enough to check the contents of the ServiceAccount just once because we know that the ServiceAccount is created atomically.
To test the stability of `sn1/test` seems excess for me.

You can check the contents even inside `Eventually()` as done in the previous `It()`.

morimoto-cybozu (changes requested):

go.mod
```diff
@@ -2,29 +2,19 @@ module github.com/cybozu-go/accurate
 
 go 1.17
```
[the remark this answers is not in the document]
masa213f: We don't use new features introduced in go 1.18, so we need not update it.

e2e/e2e_test.go
masa213f: > This Consistently() can fail if the creation of sn1/test ServiceAccount is delayed.

Thank you. That's true. If the test machine has low spec, this test may be fail.
I'll use `Eventually()`.

morimoto-cybozu (approved):
LGTM

kmdkuk (approved):
LGTM

PR #69, opened 2023-07-03
Improve release workflow

zoetrope:
The current release workflow has many manual steps.

This PR make the release workflow easier.
Just push a tag, version number will be embedded, release notes will be created, container image and Helm chart will be released.


zeroalphat (approved):
LGTM

PR #83, opened 2023-09-23
fix: add missing spec.issuerRef.group to cert-manager Certificate|

erikgb:
We are using cert-manager/approver-policy in our clusters, and are hitting this issue: https://github.com/cert-manager/approver-policy/issues/179

This PR adds the `cert-manager.io` as the issuer group - which is anyway the default in cert-manager controller.


ymmt2005 (approved):
LGTM. Thank you!

erikgb:
The CI seems to fail with an unrelated error. Is it a flake?

ymmt2005:
Looks pretty much like a flake. Re-running it.
https://github.com/cybozu-go/accurate/blob/5cf897ada5b936f579c3c578d2520e9e75796d26/hooks/namespace_test.go#L438

ymmt2005:
passed.

PR #89, opened 2023-09-27
Make admin privileges configurable

### LINKED ISSUE #82: Controller RBAC is too permissive
**Describe the bug**

The cluster role bound to the controller is too permissive by default, and I think this is a bug: https://github.com/cybozu-go/accurate/blob/main/charts/accurate/templates/generated/generated.yaml#L77-L84

If this permissive RBAC is required for the controller to operate, I think **why** should be documented.

We are evaluating this project as an alternative to HNC and might file a few issues/PRs for minor fixes. I hope there is a maintainer team with some bandwidth and interest in "external" contributions. 😄

ymmt2005:
Fix #82.

With this change, we stop granting the below permission to the
accurate controller.

```yaml
  - apiGroups:
      - '*'
    resources:
      - '*'
    verbs:
      - get
      - list
      - watch
```

Also, we make the ClusterRole `admin` optional.
The Helm chart now takes optional ClusterRoles to be granted.


README.md
```diff
@@ -49,7 +49,11 @@ Run and try Accurate on a [kind (Kubernetes-In-Docker)][kind] cluster as follows
 
 3. Install [aqua][].
 
-    https://aquaproj.github.io/docs/tutorial-basics/quick-start
```
ymmt2005, line 52: This page no longer exists.

morimoto-cybozu:
@ymmt2005 
Don't we need to remove `//+kubebuilder:rbac:...` from `controllers/propagate.go`?

charts/accurate/values.yaml
```diff
@@ -102,3 +102,8 @@ controller:
           - update
           - patch
           - delete
+    # controller.additionalRBAC.clusterRoles -- Specify additional ClusterRoles to be granted
+    # to the accurate controller. "admin" is recommended to allow the controller to manage
+    # common namespace-scoped resources.
+    clusterRoles:
+    - admin
```
erikgb, line 109:
```suggestion
      - admin
```

Nit

erikgb (changes requested):
The changes look good to me, but I think this should be removed: https://github.com/cybozu-go/accurate/blob/main/controllers/propagate.go#L57, and then run controller-gen.

charts/accurate/values.yaml
ymmt2005: Thanks. But I don't take formatting preference unless we enforce it by checking it on CI.

erikgb: I just think we should have consistent formatting in the same file. 😉

morimoto-cybozu (approved):
LGTM

Issue #87, opened 2023-09-25
Should not attempt to modify resources in normal namespace


erikgb:
**Describe the bug**

We are running a proof-of-concept in one of our OpenShift clusters. And while the controller runs find without any watches, it fails badly when we enable watches of roles and role bindings. This issue might be relevant for other resource types, but namespace RBAC is the obvious start for propagating namespace resources for us.

It seems like the controller attempts to annotate resources in namespaces that are not annotated with any `accurate.cybozu.com` annotations, and this must be wrong! Errors are emitted constantly, and here is an example:

````json
{"level":"error","ts":"2023-09-25T18:21:37Z","msg":"failed to check the controller reference","controller":"role","controllerGroup":"rbac.authorization.k8s.io","controllerKind":"Role","Role":{"name":"cluster-samples-operator","namespace":"openshift-cluster-samples-operator"},"namespace":"openshift-cluster-samples-operator","name":"cluster-samples-operator","reconcileID":"957d956a-b15e-4463-a458-ca5b1a32582b","error":"failed to add accurate.cybozu.com/propagate-generated annotation: roles.rbac.authorization.k8s.io \"cluster-samples-operator\" is forbidden: user \"system:serviceaccount:accurate:accurate-controller-manager\" (groups=[\"system:serviceaccounts\" \"system:serviceaccounts:accurate\" \"system:authenticated\"]) is attempting to grant RBAC permissions not currently held"
````

This is a total blocker for us. The controller must **NOT** touch resources in namespaces that are irrelevant for Accurate.

**Environments**
- Version: 1.1.0
- OS: N/A

**To Reproduce**

1. Install Accurate
2. Configure Accurate to watch roles and role bindings
3. Observe that Accurate adds (or attempts to add) annotations to resources in namespaces that are not observed/managed by Accurate. On Openshift this update is unsuccessful and the controller logs become extremely chatty.

**Expected behavior**

No attempt to update irrelevant namespace resources and no errors in controller logs.

**Additional context**

The controller errors might only occur on Openshift, which is a secure by default Kubernetes distribution. But the root cause here is the wrong attempt to change irrelevant resources.

zeroalphat:
@erikgb 
Thanks for your feedback.
We will investigate it.

ymmt2005:
This is fixable but not easy.
The accurate controller tries to add that annotation to record if the resource is generated from another resource, mainly for performance reasons.

[...]

PR #92, opened 2023-10-03
feat: add DisablePropagateGenerated feature gate

erikgb:
This adds a temporary/experimental feature gate that allows a user to disable the [Annotating a resource to propagate resources created from it](https://cybozu-go.github.io/accurate/propagation.html#annotating-a-resource-to-propagate-resources-created-from-it) feature. As the Accurate controller works at present, we need a way to prevent Accurate from touching resources in namespaces that are not and never will be, managed by Accurate.

The idea is to deprecate and eventually remove this new feature gate when the root cause is fixed.

Relates to https://github.com/cybozu-go/accurate/issues/87 (workaround)

````
$ bin/accurate-controller --help
accurate controller

Usage:
  accurate-controller [flags]

Flags:
      --add_dir_header                    If true, adds the file directory to the header of the log messages
      --alsologtostderr                   log to standard error as well as files (no effect when -logtostderr=true)
      --apiserver-qps-throttle int        The maximum QPS to the API server. (default 50)
      --cert-dir string                   webhook certificate directory
      --config-file string                Configuration file path (default "/etc/accurate/config.yaml")
      --feature-gates mapStringBool       A set of key=value pairs that describe feature gates for alpha/experimental features. Options are:
                                          AllAlpha=true|false (ALPHA - default=false)
                                          AllBeta=true|false (BETA - default=false)
                                          DisablePropagateGenerated=true|false (ALPHA - default=false)

````

### DIFF
--- a/cmd/accurate-controller/sub/root.go
+++ b/cmd/accurate-controller/sub/root.go
@@ -9,6 +9,7 @@ import (
 	"strconv"
 
 	"github.com/cybozu-go/accurate"
+	"github.com/cybozu-go/accurate/pkg/config"
 	"github.com/spf13/cobra"
 	klog "k8s.io/klog/v2"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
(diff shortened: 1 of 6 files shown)


ymmt2005 (approved):
LGTM.
@zoetrope could you take another look?

zoetrope (approved):
LGTM

Issue #85, opened 2023-09-25
SubNamespace should be compatible with kstatus


erikgb:
## What

We use FluxCD to provision our clusters, and FluxCD supports [health checks](https://fluxcd.io/flux/components/kustomize/kustomizations/#health-checks) for CRDs compatible with [kstatus](https://github.com/kubernetes-sigs/cli-utils/blob/master/pkg/kstatus/README.md). And I believe ArgoCD also does this.

It would be nice if the SubNamespace CRD had a status subresource compatible with kstatus. You can read more details in the specification, but I would suggest the following requirements (at least as a start):

- A SubNamespace with empty status has not yet been picked up by the controller
- The controller should ensure `.status.obeservedGeneration` is equal to `.metadata.generation` when reconciling a SubNamespace
- If the actual state differs from the desired state, the controller should add status conditions indicating that a reconcile is in progress
- When the actual state equals the desired status, the controller should remove all status conditions
- If the controller is unable to reconcile successfully for any reason, like in a conflict situation, the controller should add conditions indicating what's wrong.

Since the `status` field on SubNamespace currently is a simple string, this will be a **breaking change of status**.

## How

Describe how to address the issue.

## Checklist

- [ ] Finish implementation of the issue
- [ ] Test all functions
- [ ] Have enough logs to trace activities
- [ ] Notify developers of necessary actions

zeroalphat:
@erikgb 
Thanks for your feedback.
We will investigate it.

ymmt2005:
To address this, we need to make an incompatible change to the API.
I mean, we need to cut `v2`.

PR #96, opened 2023-10-14
chore: change v2alpha1 SubNamespace status to kstatus-compatible subresource

erikgb:
A new and currently unserved API version was added with https://github.com/cybozu-go/accurate/pull/101. This PR makes `status` in the new version a subresource compatible with [kstatus](https://github.com/kubernetes-sigs/cli-utils/blob/master/pkg/kstatus/README.md).

I have also added conversion code with unit tests, so you can get an idea of how I plan to move forward after this is (hopefully) merged. The next steps will probably be:

1. Add conversion webhook using the conversion logic established in this PR.
2. Make `v2alpha1` served, migrate controller to reconcile `v2alpha1` and make `v2alpha1` the storage version.

I would prefer using SSA when updating `status` from controller in the latter step. Please let me know what you think! This is probably best done by generating apply-configurations in a separate pre-PR.

Relates to  https://github.com/cybozu-go/accurate/issues/85

NOTE: The proposed fuzz conversion tests added in this PR are highly inspired by the equivalent test setup in https://github.com/kubernetes-sigs/cluster-api. Example: https://github.com/kubernetes-sigs/cluster-api/blob/main/api/v1alpha4/conversion_test.go


erikgb:
If you want, I can split this PR:

1. Establish the new API version as a copy of v1
2. Make the new version compatible with kstatus

erikgb:
I have now created a dedicated PR for (just) introducing the new API version: https://github.com/cybozu-go/accurate/pull/101

erikgb:
@zoetrope @ymmt2005 I think this PR is ready for a round of review now. Please let me know what you think!

zoetrope (approved):
Thank you for the excellent pull request!
I believe the PR looks great.

I wasn't familiar with the "kstatus" library before, but it looks interesting. 
In situations like this, I often use [apimachinery's Condition](https://github.com/kubernetes/apimachinery/blob/master/pkg/api/meta/conditions.go).

erikgb:
> Thank you for the excellent pull request! I believe the PR looks great.
> 
> I wasn't familiar with the "kstatus" library before, but it looks interesting. In situations like this, I often use [apimachinery's Condition](https://github.com/kubernetes/apimachinery/blob/master/pkg/api/meta/conditions.go).

Thanks for the nice feedback, @zoetrope! We are also heavy users of [apimachinery's Condition](https://github.com/kubernetes/apimachinery/blob/master/pkg/api/meta/conditions.go) in our operators not yet migrated to SSA. I think the utilities in this package are mostly useful with CSA. But we will get to that in the follow-up PRs. 😉

erikgb:
Is there anything more needed to get this merged?

PR #99, opened 2023-10-14
chore: enable DisablePropagateGenerated feature gate by default

erikgb:
As agreed in https://github.com/cybozu-go/accurate/issues/87#issuecomment-1760038161, this graduates the `DisablePropagateGenerated ``to Beta and enables it by default. This does not fix https://github.com/cybozu-go/accurate/issues/87, but I still think we can close the issue as "not a problem" - since the problematic feature is no longer enabled by default.

Should probably be merged/released after https://github.com/cybozu-go/accurate/pull/95. This PR is quite easy to rebase, so it can just stay open until we are ready for these changes.

Close https://github.com/cybozu-go/accurate/issues/87

PS: The structure/wording of the new feature gates doc paragraph is almost a copy of https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/. Kudos to Kubernetes documentation contributors/maintainers. 🙏

PR #100, opened 2023-10-15
fix(helm): controller.extraArgs should be treated as list

erikgb:
In https://github.com/cybozu-go/accurate/pull/99 I discovered that the `config.extraArgs` Helm value is not treated as a list when rendering the Deployment template - which makes it harder to use more than one extra argument to the workload. This PR fixes this.


zoetrope (approved):
LGTM

PR #101, opened 2023-10-15
chore: introduce new (identical) SubNamespace API version

erikgb:
As suggested in https://github.com/cybozu-go/accurate/pull/96, I think it makes a lot of sense to first introduce an identical and unserved API version. Since there must be only one storage version of a CRD in Kubernetes, the current `v1` version is marked as the storage version.


api/v2alpha1/subnamespace_types.go
```diff
@@ -0,0 +1,59 @@
...
+	Labels map[string]string `json:"labels,omitempty"`
+
+	// Annotations are the annotations to be propagated to the sub-namespace.
+	// +optional
+	Annotations map[string]string `json:"annotations,omitempty"`
+}
+
+// Keeping this version un-served for now
+//+kubebuilder:unservedversion
+//+kubebuilder:object:root=true
```
ymmt2005, line 32: I guess we're going to make the `status` a subresource.
```suggestion
// Keeping this version un-served for now
// +kubebuilder:unservedversion
// +kubebuilder:object:root=true
// +kubebuilder:subresource:status
```

ymmt2005 (changes requested):

api/v2alpha1/subnamespace_types.go
erikgb: I was planning to start with an _identical_ new version, but we are heading for a status subresource. Do you prefer heading for https://github.com/cybozu-go/accurate/pull/96 directly?

ymmt2005: I understand.

ymmt2005 (approved):
LGTM

PR #106, opened 2023-11-11
chore: generate applyconfigurations for APIs (SSA)

erikgb:
This PR adds generation of applyconfigurations to facilitate type-safe SSA for custom resources. I have made the generated types internal for now, and only for the newer version of our API. The upstream generator is built for Kubernetes, requiring minor tweaks to the project layout.

I plan to migrate the SubNamespace controller to SSA when introducing the new API version (`v2alpha1`).

Please take a look and tell me what you think! I'll be happy to reduce the number of changes in this PR.

CC @zoetrope @ymmt2005


zoetrope:
@erikgb 
I think introducing SSA is a great idea.
However, when we update SubNamespace, we only update the status.
Do you think there is an opportunity to use the generated ApplyConfiguration?

erikgb:
> Do you think there is an opportunity to use the generated ApplyConfiguration?

Oh yes, that is my plan in the next PR.

Makefile
```diff
@@ -109,6 +123,13 @@ $(SETUP_ENVTEST):
 	# see https://github.com/kubernetes-sigs/controller-runtime/tree/master/tools/setup-envtest
 	GOBIN=$(shell pwd)/bin go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest
 
+APPLYCONFIGURATION_GEN := $(shell pwd)/bin/applyconfiguration-gen
+.PHONY: applyconfiguration-gen
+applyconfiguration-gen: $(APPLYCONFIGURATION_GEN) ## Download applyconfiguration-gen locally if necessary
+$(APPLYCONFIGURATION_GEN):
+	# see https://github.com/kubernetes/code-generator/tree/master/cmd/applyconfiguration-gen
+	GOBIN=$(shell pwd)/bin go install k8s.io/code-generator/cmd/applyconfiguration-gen@v0.28.3
```
zoetrope, line 131: We are using `aqua` to manage packages for `accurate` to automate package updates using `renovate`.
We would like to manage `applyconfiguration-gen` with `aqua` as well.
But that is a bit complicated. I will create a PR for that later.

zoetrope (approved):
LGTM

Makefile
zoetrope: #109

PR #107, opened 2023-11-12
feat: enable v2alpha1 API with conversion webhook

erikgb:
This PR makes the new v2alpha1 version of the API served and enables the conversion webhook using logic established in https://github.com/cybozu-go/accurate/pull/96.

After this, I will suggest changing the SubNamespace controller to watch/reconcile version v1alpha1 (using SSA). When the controller watches v2alpha1 version, I think it also makes sense to switch the stored version.

### DIFF
--- a/api/accurate/v1/subnamespace_conversion.go
+++ b/api/accurate/v1/subnamespace_conversion.go
@@ -21,7 +21,7 @@ func (src *SubNamespace) ConvertTo(dstRaw conversion.Hub) error {
 
 	logger := getConversionLogger(src).WithValues(
 		"source", SchemeGroupVersion.Version,
-		"destination", SchemeGroupVersion.Version,
+		"destination", accuratev2alpha1.SchemeGroupVersion.Version,
 	)
 	logger.V(5).Info("converting")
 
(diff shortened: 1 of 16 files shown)


erikgb:
@zoetrope @ymmt2005 Do you have any suggestions on how to solve the CRD issue in the Helm chart? It won't work as it stands now, and an additional problem is that Helm is unable to template CRD. For this reason, several open-source projects have moved the CRDs out of the Helm crds folder. And provides a Helm value to control if CRDs are provisioned or not. Helm has never had proper support for CRDs, and I think they have given up now. Please take a look at how cert-manager, which I know pretty well "solves" the Helm issue: https://cert-manager.io/docs/installation/helm/#3-install-customresourcedefinitions

If you think that's a good approach, I can prepare a preparation PR making the CRDs possible to template. And then I can progress on this PR.

zoetrope:
@erikgb 
I had a look at the cert-manager technique.
https://github.com/cert-manager/cert-manager/blob/d2f6bbe579fd9d5f88b82f9a4bfe9241709e9eb8/make/manifests.mk#L107-L110

I agree that it seems like a very good approach.
Let's proceed with this method.

erikgb:
> I agree that it seems like a very good approach. Let's proceed with this method.

Thanks for the feedback. I will prepare a PR.

erikgb:
@zoetrope I think this PR is ready for review now. I have also added a simple e2e-test to check conversion works. Next PR will be a lot more interesting - I think. 😸

zoetrope (approved):
LGTM

PR #109, opened 2023-11-15
Setup applyconfiguration-gen with aqua

zoetrope:


Makefile
```diff
@@ -115,6 +115,7 @@ release-build: setup
 
 setup:
 	aqua i -l
+	aqua policy allow ./aqua-policy.yaml
```
ymmt2005, line 118: Not sure, but doesn't this need to come before `aqua i -l`?

zoetrope: Thanks!
That's right. I don't know why but CI passed 🤔

ymmt2005: @zoetrope 
My guess is:
1. `aqua -i l` installs only links to a hook script.
2. When the CI actually runs a command installed by Aqua, the hook script tries to install the real command.
3. At that time, `aqua policy allow` has been run, so the hook script succeeds.

ymmt2005: Also note that we use `aqua-installer` action that runs `aqua -i l` by default.
https://aquaproj.github.io/docs/products/aqua-installer/#optional-inputs

ymmt2005 (approved):
LGTM

Issue #98, opened 2023-10-14
Should clean up previously propagated namespace labels/annotations


erikgb:
## What

Ref. [docs](https://cybozu-go.github.io/accurate/subnamespaces.html#setting-a-namespace-as-a-root-namespace):

> Accurate currently does not delete previously propagated labels when deleted from the parent namespace to prevent unintended deletions. Users are expected to manually delete labels/annotations that are no longer needed.

I wonder which kind of "unintended deletions" are referred to here. I would expect Accurate to clean up a previously propagated namespace label/annotation (based on Kubernetes `managedFields` ownership) when the desired state of parent/template NS (or SubNamespace) indicates that a label/annotation should not be propagated.

Could a migration to SSA (Server Side Apply) aid in this? Since controller-runtime still has limited support for SSA, ref. https://github.com/kubernetes-sigs/controller-runtime/issues/347, we probably must migrate to `Unstructured` instead of structured types. But IMO I think it would be worth it.

## How

Describe how to address the issue.

## Checklist

- [ ] Finish implementation of the issue
- [ ] Test all functions
- [ ] Have enough logs to trace activities
- [ ] Notify developers of necessary actions

ymmt2005:
> Could a migration to SSA (Server Side Apply) aid in this?

I think so.

PR #111, opened 2023-11-21
feat: make SubNamespace controller reconcile v2alpha1 version

erikgb:
This PR changes the controllers to reconcile `SubNamespace` v1alpha2 version and makes the newer version the stored version in the CRD. The modified reconciler uses SSA to update `SubNamespace` using apply configurations generated in https://github.com/cybozu-go/accurate/pull/106. I also plan to finalize migration to SSA for other resource types in use by controller.

Fixes https://github.com/cybozu-go/accurate/issues/85

We should probably migrate the webhooks and kubectl-plugin to the newer version now. I can fix the hooks, but I have no experience with kubectl-plugins, so maybe you can assist in the migration?


zoetrope:
I will check to see if the plugin needs to be fixed.

zoetrope:
@erikgb 
Apologies for the delayed response.

I've looked into the kubectl-plugins.
It appears that the plugin does not handle the Status in SubNamespace.
Additionally, since webhook will convert the resource, I see no need to change the implementation.
However, if there's a need for changes, please let me know, and I'll be happy to assist with the required updates.

erikgb:
> @erikgb Apologies for the delayed response.

@zoetrope No worries! Accurate works well for us. We just want to support kstatus to get better user feedback when using tools like FluxCD or ArgoCD.

> I've looked into the kubectl-plugins. It appears that the plugin does not handle the Status in SubNamespace. Additionally, since webhook will convert the resource, I see no need to change the implementation. However, if there's a need for changes, please let me know, and I'll be happy to assist with the required updates.

Thanks! If the plugin doesn't use status now, I think this change won't be problematic. At some point, we should probably deprecate the v1 API as a hint to users about migrating to the newer version. But I am hoping to fully migrate to SSA, which I hope can solve issues like https://github.com/cybozu-go/accurate/issues/98.

zoetrope (approved):
LGTM

PR #112, opened 2023-12-08
chore: migrate controllers to SSA

### LINKED ISSUE #98: Should clean up previously propagated namespace labels/annotations

erikgb:
This PR migrates all updates in controllers to server-side apply (SSA), except the update removing the finalizer on SubNamespace - which is changed to a standard patch.

SSA works a bit differently than updates, so some of the optimizations had to go in this PR. I am hoping this change won't increase the load on the api-server significantly, but it's something we should monitor. I have added stable resource version tests for namespaces - which are now SSAed unconditionally.

Namespace controller tests are currently highly scenario-based, which makes them long and hard to read/modify. I have added some additional assertions of modified behavior on migration to SSA, but I suggest refactoring the tests in a follow-up PR to improve the test coverage. I can prepare a PR if you agree.

Fixes #98

### DIFF
--- a/controllers/namespace_controller_test.go
+++ b/controllers/namespace_controller_test.go
@@ -201,6 +201,12 @@ var _ = Describe("Namespace controller", func() {
 
 		Eventually(komega.Object(ns1)).Should(HaveField("Labels", HaveKeyWithValue("foo.bar/baz", "123")))
 
+		By("removing a label of template namespace")
+		Expect(komega.Update(tmpl, func() {
+			delete(tmpl.Labels, "foo.bar/baz")
+		})()).To(Succeed())
+		Eventually(komega.Object(ns1)).Should(HaveField("Labels", Not(HaveKey("foo.bar/baz"))))
+
 		tmpl2 := &corev1.Namespace{}
 		tmpl2.Name = "tmpl2"
 		tmpl2.Labels = map[string]string{
@@ -229,12 +235,17 @@ var _ = Describe("Namespace controller", func() {
 		Eventually(komega.Get(pSec2)).Should(Succeed())
 
 		Eventually(komega.Object(ns1)).Should(HaveField("Labels", HaveKeyWithValue("team", "maneki")))
+		// assert that annotations/labels from previous template are removed
+		Expect(ns1.Annotations).To(Not(HaveKey("foo.bar/zot")))
+		Expect(ns1.Annotations).To(Not(HaveKey("memo")))
+		Expect(ns1.Annotations).To(Not(HaveKey("team")))
+		Expect(ns1.Labels).To(Not(HaveKey("foo.bar/baz")))
+		Expect(ns1.Labels).To(Not(HaveKey("memo")))
 
 		By("unsetting the template")
 		Expect(komega.Update(ns1, func() {
 			delete(ns1.Labels, constants.LabelTemplate)
 		})()).To(Succeed())
-
 		Eventually(komega.Get(pSec2)).Should(WithTransform(apierrors.IsNotFound, BeTrue()))
 	})
 
@@ -278,6 +289,12 @@ var _ = Describe("Namespace controller", func() {
 		Expect(instance.Annotations["memo"]).Should(Equal("test"))
 
 		Expect(komega.Object(tmpl2)()).To(HaveField("Labels", HaveKeyWithValue("team", "neco")))
+
+		By("removing a label from root template namespace")
+		Expect(komega.Update(tmpl1, func() {
+			delete(tmpl1.Labels, "team")
+		})()).To(Succeed())
+		Eventually(komega.Object(instance)).Should(HaveField("Labels", Not(HaveKey("team"))))
 	})
 
 	It("should not delete resources in an independent namespace", func() {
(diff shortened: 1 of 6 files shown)


erikgb:
After testing this PR I think it needs a bit more work - especially to avoid endless reconcile loops.

erikgb:
I think this PR is ready for review now. After we hopefully get this in, I am hoping to get a new release soon. 🙏 There are already many interesting unreleased improvements, and fixing #98 (this PR) is a high priority for us.

While you review this PR, I will prepare some vanilla release preparation PRs like dumping dependencies etc. Have you ever considered enabling Depandabot/Renovate on this project?

yamatcha:
@erikgb 

> Namespace controller tests are currently highly scenario-based, which makes them long and hard to read/modify. I have added some additional assertions of modified behavior on migration to SSA, but I suggest refactoring the tests in a follow-up PR to improve the test coverage. I can prepare a PR if you agree.

Current test is the scenario we need in our environment, so we don't want to lose some of them in the major refactoring.
Would you tell me how you want to refactor this.

docs/subnamespaces.md
```diff
@@ -30,7 +30,7 @@ metadata:
     team: foo
 ```
 
-Accurate only propagates labels/annotations that have been configured in that respect via the `labelKeys` and `annotationKeys` parameters in `config.yaml`. This prevents the propagation of labels/annotations that were not meant to do so. Accurate currently does not delete previously propagated labels when deleted from the parent namespace to prevent unintended deletions. Users are expected to manually delete labels/annotations that are no longer needed.
+Accurate only propagates labels/annotations that have been configured in that respect via the `labelKeys` and `annotationKeys` parameters in `config.yaml`. This prevents the propagation of labels/annotations that were not meant to do so.
```
yamatcha, line 33: This PR includes a change in specifications.
Could you please tell me why you are changing this behavior?

ymmt2005: @yamatcha 
This is the very reason why @erikgb is making this PR.
Without SSA, it was difficult to remove labels and annotations safely.
Now with SSA, removing them is a safe operation.

erikgb: Exactly! @ymmt2005 is correct that this is the main motivation for this PR. There is yet no 1. class support for SSA in controller-runtime, but I have been involved in migrating several operators to SSA using the same approach as in this PR. So I am pretty sure it works, and it's also properly tested now IMO. The main concern with SSA migration is the risk of getting never-ending reconcile loops, but I have added new tests for this.

We would be extremely happy if this PR could eventually get merged and released. https://github.com/cybozu-go/accurate/issues/98 is a semi-blocker for our further rollout of Accurate in our clusters.

erikgb:
> @erikgb
> 
> > Namespace controller tests are currently highly scenario-based, which makes them long and hard to read/modify. I have added some additional assertions of modified behavior on migration to SSA, but I suggest refactoring the tests in a follow-up PR to improve the test coverage. I can prepare a PR if you agree.
> 
> Current test is the scenario we need in our environment, so we don't want to lose some of them in the major refactoring. Would you tell me how you want to refactor this.

This was just a suggestion and not a requirement. I have no problems leaving the tests as-is. But if I were a maintainer, I would split things more. I don't see a big benefit in the long scenario-based tests, and think the test code would be easier to read if the scenarios were tested one by one. The extensive use of Ginkgo `By` construct is a sign of this (but required for these lengthy tests). 😄 No offense, of course. This is just my point of view.

docs/subnamespaces.md
yamatcha: @erikgb 
Thank you for the detailed description! 
I missed issue refered  in PR description.

controllers/namespace_controller_test.go
```diff
@@ -462,4 +476,60 @@ var _ = Describe("Namespace controller", func() {
 			HaveField("Annotations", HaveKeyWithValue("memo", "tama")),
 		))
 	})
+
+	Context("templated namespace", func() {
```
yamatcha, line 480: What do you want this test to confirm?
I think confirming that ns is created from template and that labels and annotations are propagated has already been tested here.
https://github.com/cybozu-go/accurate/blob/cda331bc3d8386a655695191b419f22863d54db4/controllers/namespace_controller_test.go#L163

erikgb: This test is to ensure consistent reconciles. Reconcile loops is the one thing to fear when migrating to SSA.

yamatcha: Thanks to you, I understood.

yamatcha (approved):
LGTM

PR #95, opened 2023-10-14
docs: deprecate propagate-generated feature

erikgb:
As agreed in https://github.com/cybozu-go/accurate/issues/87#issuecomment-1760038161 we think it's time to initiate the removal of the propagate-generated feature. This PR does:
- update docs to inform about the deprecated feature
-  mark affected code as deprecated

Relates to https://github.com/cybozu-go/accurate/issues/87
Close https://github.com/cybozu-go/accurate/issues/93

PR #113, opened 2023-12-08
chore: deprecate user-facing propagate-generated annotation

erikgb:
This fixes a leftover after https://github.com/cybozu-go/accurate/pull/95. Somehow we forgot to deprecate the user-facing annotation for the now deprecated propagate-generated feature.


yamatcha (approved):
LGTM

Issue #120, opened 2024-02-23
Pre-existing resources should be upgraded to SSA


erikgb:
**Describe the bug**

We have installed the latest version of Accurate (1.3.0), which includes migration to SSA (https://github.com/cybozu-go/accurate/pull/112). While everything seems to work well, we have noticed that pre-existing resources (from before we installed the release with migration to SSA) have some suspicious `managedFields`. Example for a namespace controlled by a sub-namespace:

````yaml
managedFields:
  - manager: accurate-controller
    operation: Apply
    apiVersion: v1
    fieldsType: FieldsV1
    fieldsV1:
      'f:metadata':
        'f:annotations':
          'f:collectord.io/index': {}
        'f:labels':
          'f:trust.statnett.no/inject-statnett-bundle': {}
  - manager: accurate-controller
    operation: Update
    apiVersion: v1
    time: '2023-11-08T15:01:07Z'
    fieldsType: FieldsV1
    fieldsV1:
      'f:metadata':
        'f:annotations':
          .: {}
          'f:collectord.io/index': {}
        'f:labels':
          .: {}
          'f:accurate.cybozu.com/parent': {}
          'f:app.kubernetes.io/created-by': {}
          'f:kubernetes.io/metadata.name': {}
          'f:trust.statnett.no/inject-statnett-bundle': {}
````

This results in unset fields in the SSA patch to retain in the owned resource, which is the issue the migration to SSA PR was supposed to solve (https://github.com/cybozu-go/accurate/issues/98).

**Environments**

N/A

**To Reproduce**

1. Install one of the previous versions of Accurate before we migrated to SSA.
2. Create a SubNamespace adding a label/annotation to the sub-namespace.
3. Upgrade Accurate to the latest version including migration to SSA
4. Remove sub-namespace a label/annotation from `SubNamespace` created in 2.
5. Observe the label/annotation is still present on sub-namespace.

**Expected behavior**

The label/annotation removed from `SubNamespace` spec should be removed from sub-namespace.

**Additional context**

I know how to fix this and will prepare a PR. 🤠

PR #121, opened 2024-02-23
fix: upgrade managed fields in controller resources from CSA to SSA

erikgb:
This PR reuses the upstream code that `kubectl` uses to upgrade `managedFields` from CSA to SSA. It appears like subresources are not supported yet, but I am working on it in https://github.com/kubernetes/kubernetes/pull/123484. The PR allowing us to upgrade managed fields for subresources is now merged and should be released with the upcoming K8s 1.30. I will create a follow-up PR when the improved helper is available, but this PR fixes the main issue.

Fixes https://github.com/cybozu-go/accurate/issues/120

### DIFF
--- a/controllers/namespace_controller.go
+++ b/controllers/namespace_controller.go
@@ -118,6 +118,15 @@ func (r *NamespaceReconciler) propagateMeta(ctx context.Context, ns, parent *cor
 				}
 			}
 		}
+		// Must ensure we set all fields we care for, also labels added when creating namespace
+		labels[constants.LabelCreatedBy] = constants.CreatedBy
+		labels[constants.LabelParent] = parent.Name
+	}
+
+	// Ensure that managed fields are upgraded to SSA before the following SSA.
+	// TODO(migration): This code could be removed after a couple of releases.
+	if err := upgradeManagedFields(ctx, r.Client, ns); err != nil {
+		return err
 	}
 
 	ac := corev1ac.Namespace(ns.Name).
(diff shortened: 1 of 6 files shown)


erikgb:
@ymmt2005, do you have any suggestions on how to test this properly? Some of the owned resources are created with a CREATE request, which will set the `managedFields` to Update (CSA). I could test by forcing an update of the owned resource with SSA - which should also upgrade the resource from CSA to SSA. But I was hoping to replace the initial CREATE with an SSA to avoid wasting api-server resources, so this will only be a temporary test-solution. 🤔 What I want to do, is to pre-create a resource simulating a resource that existed before migrating to SSA. Could it be an option to do it in the controller's test-suite setup - before the controllers/manager are bootstrapped? WDYT?

ymmt2005:
@erikgb 
Hello,
Appreciate your continued contribution!

> What I want to do, is to pre-create a resource simulating a resource that existed before migrating to SSA. Could it be an option to do it in the controller's test-suite setup - before the controllers/manager are bootstrapped?

I'm OK with this.

erikgb:
@ymmt2005 Do you know what's wrong with the CI? Seems unrelated to the change here, and I cannot reproduce it locally. It appears to complain about some aqua "policies".

yamatcha:
@erikgb  
Most of this error log is unimportant what is important is this the following part.
https://github.com/cybozu-go/accurate/actions/runs/8030385773/job/21937711586?pr=121#step:6:38
I think CI failed because `my secret` was not found here.
https://github.com/cybozu-go/accurate/blob/36bb411afb82b1f56ad0767379c5bd3c130619c9/e2e/e2e_test.go#L174

The warning log by aqua may confuse developers, so I'll fix it in another PR.

erikgb:
Thanks for the feedback @yamatcha! I will take another look!

erikgb:
I finally found the reason for the failing test. When using SSA you need to set ALL fields the manager cares for. These two lines were missing:

https://github.com/cybozu-go/accurate/commit/e899d8797aba131a4b5a398490d134ee202f129d#diff-33975f8c70294bffa37f57ef00ebd5656b884bad2006be7e261ef8cca1011161R122-R123

No idea why the envtests didn't catch this.... PTAL!

yamatcha (approved):

PR #128, opened 2024-05-15
chore: deprecate v1 and v2alpha1 SubNamespace API

erikgb:
The v2alpha1 version of the SubNamespace API has been available for a while, and I think it's time to consider deprecating the v1 version. The newer version has a much better observed state UX with conditions. We should probably make v2 (or v2beta1) available - to not "scare" our users from migrating (to an alpha version).

PR #132, opened 2024-06-07
feat: start serving v2 API

erikgb:
With this PR, we'll start serving the new v2 API - which is identical to v2alpha1 API. Relates to https://github.com/cybozu-go/accurate/pull/128.

I'll prepare another PR to switch the stored version from v2alpha1 to v2 in a follow-up PR.

NOTE: It seems like the v2 API now becomes the new default API version when using kubectl without specifying the version of `SubNamespace`.


zoetrope (approved):
LGTM

Issue #102, opened 2023-10-19
Operator should crash if missing RBAC to any watched resource


erikgb:
## What

After adding a new watched resource type, we forgot to grant the controller RBAC to operator on the new resources. This made the operator workload roll over apparently with success. But after checking the logs, I see that it emits tons of errors. Example:

````
{"level":"error","ts":"2023-10-19T11:29:09Z","msg":"Reconciler error","controller":"namespace","controllerGroup":"","controllerKind":"Namespace","Namespace":{"name":"bsrv"},"namespace":"","name":"bsrv","reconcileID":"d8c88184-5aa8-4de9-961c-c57776f6bed0","error":"failed to reconcile a namespace: failed to propagate resource bsrv/default-limits of /v1, Kind=LimitRange with propagate=create: limitranges is forbidden: User \"system:serviceaccount:accurate:accurate-controller-manager\" cannot create resource \"limitranges\" in API group \"\" in the namespace \"bsrv\"","stacktrace":"sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.16.2/pkg/internal/controller/controller.go:329\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.16.2/pkg/internal/controller/controller.go:266\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller...
````

## How

I think the controller should perform a check on startup, to see if it has the required RBAC to the configured watched resources, and crash/panic if it doesn't. The check could be based on the [SelfSubjectAccessReview](https://kubernetes.io/docs/reference/kubernetes-api/authorization-resources/self-subject-access-review-v1/) API.

## Checklist

- [ ] Finish implementation of the issue
- [ ] Test all functions
- [ ] Have enough logs to trace activities
- [ ] Notify developers of necessary actions

PR #134, opened 2024-06-08
fix: check RBAC to watched resources on startup

erikgb:
This PR adds more validation of configuration on startup. The operator must have cluster-wide RBAC to CRUD all configured watches to function as intended by the user. At startup, we will now use [SelfSubjectAccessReview](https://kubernetes.io/docs/reference/kubernetes-api/authorization-resources/self-subject-access-review-v1/) to check the required RBAC for all configured watches and crash if anything is missing.

Note: After migrating to SSA, we no longer need the `update` verb. I included this minor change in this PR, but please let me know if a dedicated PR is required.

Close https://github.com/cybozu-go/accurate/issues/102

### DIFF
--- a/charts/accurate/MIGRATION.md
+++ b/charts/accurate/MIGRATION.md
@@ -114,7 +114,6 @@ controller:
           - list
           - watch
           - create
-          - update
           - patch
           - delete
 <snip>
--- a/charts/accurate/values.yaml
+++ b/charts/accurate/values.yaml
@@ -101,7 +101,6 @@ controller:
           - list
           - watch
           - create
-          - update
           - patch
           - delete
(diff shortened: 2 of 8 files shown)


zoetrope (commented):
Thanks for the PR!
Would it be difficult to add tests for this feature?

erikgb:
> Thanks for the PR! Would it be difficult to add tests for this feature?

I thought about tests when preparing the PR. I think it's a bit more difficult than usual, but I agree there should ideally be tests for everything. I will take another look!

erikgb:
I have now added some initial tests to this feature. Please let me know if more detailed testing is required.

zoetrope (approved):

zoetrope:
@erikgb 
Thanks!

PR #137, opened 2024-06-15
test: migrate tests to v2 API

erikgb:
After start serving v2 API it makes sense to migrate tests to this new version. Older API versions will hopefully be deprecated by merging https://github.com/cybozu-go/accurate/pull/128.

After this PR, we have little use of older API versions internal. @zoetrope, maybe you can migrate the kubectl plugin to use v2 API?

### DIFF
--- a/e2e/e2e_test.go
+++ b/e2e/e2e_test.go
@@ -6,11 +6,12 @@ import (
 	"errors"
 	"os"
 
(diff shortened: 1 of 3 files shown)


zoetrope (approved):
Thanks, I will check on kubectl support.

Issue #53, opened 2022-05-12
Exclude particular labels/annotations from propagation


ymmt2005:
## What

Accurate can propagate any namespace resource between namespaces.
While copying the resource, all labels and annotations except for ones that contain `kintone.io/` are inherited.
https://github.com/cybozu-go/accurate/blob/ae009976539579320acf5765081c0590ca93819d/controllers/propagate.go#L24-L55

In some cases, the copied labels or annotations can cause problems.
For instance, if Argo CD is configured to track the managed resources by `argocd.argoproj.io/instance` label and the parent resource was created by Argo CD, the propagated resource would have the same label. Argo CD then tries to delete the propagated resource because the propagated resource is not found on the source Git repository.

So, add a feature to exclude particular labels or annotations from propagated resources.

## How

Describe how to address the issue.

## Checklist

- [ ] Finish implementation of the issue
- [ ] Test all functions
- [ ] Have enough logs to trace activities
- [ ] Notify developers of necessary actions

erikgb:
@ymmt2005 Do you have any thoughts about the UX of this feature? Should it be configured:

1. globally on the controller level, excluding labels/annotations for **all** propagated resources
2. under watches in the controller configuration, excluding labels/annotations **per resource type** of propagated resources
3. new label/annotation allowing this to be configured **per resource instance** of propagated resources

erikgb:
@zoetrope @ymmt2005 We are facing this issue in our clusters, and I want to fix it. Do you have some input to my question above? I think the implementation complexity will increase in order from 1 to 3. I only need this configured globally, so I'll vote for 1 - at least as a start.

ymmt2005:
@erikgb Sorry, I didn't notice your question.

I agree with option 1.

PR #142, opened 2024-06-30
feat: configure excluded propagate labels/annotations

erikgb:
This PR proposes an implementation of https://github.com/cybozu-go/accurate/issues/53. I have tried to keep changes to a minimum though it seems like the code could benefit from some simple refactorings after this change.

I struggled to find good names for the new config props, so please let me know if you have better suggestions!

The change is theoretically not 100% backwards compatible, as `strings.Contains(k, "kubernetes.io/")` will potentially exclude more than `path.Match("*kubernetes.io/*", key)`, but IMO it should give the same results for typical labels, at least https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/.

Closes https://github.com/cybozu-go/accurate/issues/53


zoetrope (commented):
> I struggled to find good names for the new config props, so please let me know if you have better suggestions!

I think the current name works fine, but how about simplifying it to `excludeLabelKeys`, `excludeAnnotationKeys`?

> The change is theoretically not 100% backwards compatible, as strings.Contains(k, "kubernetes.io/") will potentially exclude more than path.Match("*kubernetes.io/*", key), but IMO it should give the same results for typical labels, at least https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/.

No problem at all.

erikgb:
> > I struggled to find good names for the new config props, so please let me know if you have better suggestions!
> 
> I think the current name works fine, but how about simplifying it to `excludeLabelKeys`, `excludeAnnotationKeys`?

I looked in that direction, but I fear some confusion with the existing `labelKeys`/`annotationKeys` which denotes labels/annotations on namespaces - which is a quite different configuration. Maybe @ymmt2005 has some input here?

ymmt2005:
@erikgb @zoetrope 
Thank you for fixing the issue.

I agree that `exclucdeLabelKeys` would be confusing, and I'm fine with Eric's suggestion.

zoetrope:
@erikgb @ymmt2005 
Alright, let's merge it as is.

zoetrope (approved):

Issue #144, opened 2024-07-16
Should not attempt to finalize SubNamespace containing sub-namespaces


erikgb:
**Describe the bug**

Today we experienced a strange issue in one of our clusters: It seems like a client managed to delete a SubNamespace that should have been blocked by the Accurate SubNamespace webhook - since the sub-namespace has child namespaces. This is probably not correct, but the result is that Accurate controller is trying to do something that is blocked by the Accurate Namespace webhook.

From controller logs:

````json
{"level":"info","ts":"2024-07-16T12:05:08Z","msg":"starting finalization","controller":"subnamespace","controllerGroup":"accurate.cybozu.com","controllerKind":"SubNamespace","SubNamespace":{"name":"sam-review","namespace":"sam"},"namespace":"sam","name":"sam-review","reconcileID":"61bee8e0-880a-4717-88d3-13d6f8834620"}
{"level":"error","ts":"2024-07-16T12:05:08Z","msg":"Reconciler error","controller":"subnamespace","controllerGroup":"accurate.cybozu.com","controllerKind":"SubNamespace","SubNamespace":{"name":"sam-review","namespace":"sam"},"namespace":"sam","name":"sam-review","reconcileID":"61bee8e0-880a-4717-88d3-13d6f8834620","error":"failed to finalize: failed to delete namespace sam-review: admission webhook \"namespace.accurate.cybozu.io\" denied the request: child namespaces exist","stacktrace":"sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.18.0/pkg/internal/controller/controller.go:324\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.18.0/pkg/internal/controller/controller.go:261\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.18.0/pkg/internal/controlle...
````

**Environments**
- Version: 1.4.1
- OS:

**To Reproduce**

Not sure if I know how to reproduce.

**Expected behavior**

[...]

PR #145, opened 2024-07-16
fix(webhook): deny deletion of SubNamespace with child namespaces

erikgb:
If a client is allowed to delete a `SubNamespace` where the sub-namespace has child namespaces, the controller becomes stuck on the finalization of the `SubNamespace`. The controller will not be allowed to delete the sub-namespaces by our namespace webhook.

Since Kubernetes has no way to undelete a resource (unset the `deletionTimestamp` field), all child namespaces must be manually deleted to allow the `SubNamespace` deletion to be finalized.

This PR introduces handling of DELETE verb to the subnamespaces validating webhook ensuring no subnamespaces can be deleted if the sub-namespace has child namespaces. It seems like there was a minor bug in the validating webhook configuration: it was configured to handle UPDATE verb, but the implementation was only addressing CREATE.

Close https://github.com/cybozu-go/accurate/issues/144

### DIFF
--- a/charts/accurate/templates/generated/generated.yaml
+++ b/charts/accurate/templates/generated/generated.yaml
@@ -297,7 +297,7 @@ webhooks:
           - v1
         operations:
           - CREATE
-          - UPDATE
+          - DELETE
         resources:
           - subnamespaces
     sideEffects: None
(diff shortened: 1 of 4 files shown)


zoetrope (approved):
LGTM

PR #146, opened 2024-07-18
chore: remove use of deprecated APIs

erikgb:
This PR removes all usage of deprecated APIs/types that can be done before we remove the deprecated APIs.


zoetrope (approved):
Thanks a lot!

PR #149, opened 2024-10-22
fix: should resolve SubNamespace conflict when sub-namespace deleted

erikgb:
#### Background

Our tenants use GitOps to provision sub-namespaces, and not `kubectl-accurate`. We run on OpenShift, and OpenShift has a feature allowing users in the `self-provisioner` cluster role to create namespaces by issuing a ProjectRequest create request.   We want to migrate to Accurate, but will still have to support the OpenShift mechanism for a long time.

#### Problem (solved in this PR)

Recently we've had users pre-creating an OpenShift Project before provisioning the _SubNamespace_ resource that is supposed to control the desired namespace. This will not work, making the subnamespace enter a conflicting state right after creation by the tenant GitOps. The user observes this problem by inspecting the _SubNamespace_ status and other GitOps failures caused by missing RBAC in the target namespace.

The natural thing to do when this happens, is to delete the OpenShift `Project` resource currently controlling the (empty) target namespace. When the project is deleted, OpenShift will terminate the controlled namespace. So far, so good.

Expected behavior: Accurate detects the namespace deletion and (re)creates the target namespace as sub-namespace, resolving the conflict state on the _SubNamespace_.

Actual behavior: Accurate eventually acts on the deleted namespace, resolving the conflict state, but this is not happening in a timely fashion. The exact time depends on the resync time initiated regularly by controller-runtime. But we have seen this take several hours, which is not acceptable.

### DIFF
--- a/cmd/accurate-controller/sub/run.go
+++ b/cmd/accurate-controller/sub/run.go
@@ -133,6 +133,9 @@ func subMain(ns, addr string, port int) error {
 	hooks.SetupNamespaceWebhook(mgr, dec)
 
 	// SubNamespace reconciler & webhook
+	if err := indexing.SetupIndexForSubNamespace(ctx, mgr); err != nil {
+		return fmt.Errorf("failed to setup indexer for subnamespaces: %w", err)
+	}
 	if err = (&controllers.SubNamespaceReconciler{
 		Client: mgr.GetClient(),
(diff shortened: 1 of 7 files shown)


zoetrope (approved):
LGTM!
This is not a problem limited to OpenShift, but a widely useful implementation.

PR #153, opened 2025-02-17
Fix N+1 Issue in Namespace Search and Enhance Tree View

012xx:
## Overview
- Resolved the N+1 problem by optimizing namespace retrieval for faster searches.
- Fixed tree display indentation using tree symbols.
## Fixes
- Optimized namespace hierarchy retrieval to eliminate redundant queries.
- Improved tree view formatting by using tree symbols and removing unnecessary indent variables.
- Sorted namespaces in ascending alphabetical order.
## Rationale
- The current implementation suffers from an N+1 problem when searching namespaces, leading to poor performance.
- This update improves user experience by significantly speeding up namespace lookups.

### DIFF
--- a/cmd/kubectl-accurate/sub/list.go
+++ b/cmd/kubectl-accurate/sub/list.go
@@ -3,6 +3,7 @@ package sub
 import (
 	"context"
 	"fmt"
+	"sort"
 
 	"github.com/cybozu-go/accurate/pkg/constants"
 	"github.com/spf13/cobra"
@@ -11,8 +12,6 @@ import (
 	"k8s.io/cli-runtime/pkg/genericiooptions"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
-
-const indent = 4
 
 type listOptions struct {
 	streams genericiooptions.IOStreams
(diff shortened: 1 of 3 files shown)


cmd/kubectl-accurate/sub/list.go
```diff
@@ -57,44 +57,58 @@ func (o *listOptions) Fill(streams genericiooptions.IOStreams, config *genericcl
...
-	fmt.Fprintf(o.streams.Out, "%*s%s%s\n", level, "", subMark, name)
+	fmt.Fprintf(o.streams.Out, "%*s%s\n", level, "", name)
 
-	children := &corev1.NamespaceList{}
-	if err := o.client.List(ctx, children, client.MatchingLabels{constants.LabelParent: name}); err != nil {
-		return fmt.Errorf("failed to list the children of %s: %w", name, err)
+	children, exists := nsMap[name]
+	if !exists || len(children) == 0 {
+		return nil
 	}
```
takahiro-yamada, line 106: With this change, `kubectl accurate list` shows not only the namespaces which labeled with the specified one but all. Please confirm which behavior is preferable.

cmd/kubectl-accurate/sub/list.go
```diff
@@ -57,44 +57,58 @@ func (o *listOptions) Fill(streams genericiooptions.IOStreams, config *genericcl
...
+	if name == "" {
+		return fmt.Errorf("invalid namespace name: empty string")
 	}
 
-	subMark := " "
-	if _, ok := ns.Labels[constants.LabelParent]; ok {
-		subMark = "⮡"
-	}
-	fmt.Fprintf(o.streams.Out, "%*s%s%s\n", level, "", subMark, name)
+	fmt.Fprintf(o.streams.Out, "%*s%s\n", level, "", name)
```
takahiro-yamada, line 101: If you don't implement an alternative sub-mark, it's seems better to keep compatibility, how about it?

takahiro-yamada (changes requested):

cmd/kubectl-accurate/sub/list.go
012xx: The previous code was designed to display only the specified namespace and the LabelParent of the specified namespace, so we modified it to match.

takahiro-yamada (approved):
LGTM

@zoetrope 
Could you review this PR as a core maintainer please?

cmd/kubectl-accurate/sub/template_list.go
```diff
@@ -55,49 +56,78 @@ func (o *templateListOpts) Fill(streams genericiooptions.IOStreams, config *gene
...
+		nsMap[ns.Name] = ns
+		if parent, ok := ns.Labels[constants.LabelTemplate]; ok {
+			childMap[parent] = append(childMap[parent], ns)
 		}
+	}
 
-		if err := o.showNS(ctx, ns.Name, 0); err != nil {
-			return err
+	for _, children := range childMap {
+		sort.Slice(children, func(i, j int) bool {
```
zoetrope, line 80: You are sorting too many times.
`allNamespaces.Items` is already sorted, there is no need to sort children here.

zoetrope (approved):
LGTM

Issue #155, opened 2025-03-19
Request - cross-platform images


jwhitaker-gridcog:
Hey guys,
Thanks for maintaining and open-sourcing this, it's working really nicely for us.
A question - would you consider building cross-platform images? We'd love to be able to have this run on arm64 nodes.

I'm about to see if I can get your Github Actions setup to do it easily, but just starting the conversation before flinging PRs around :)

erikgb:
I highly support this request, but I would also love to change the base image to something more common and transparent, ref. https://github.com/cybozu-go/accurate/issues/103. In our setup, Accurate has cluster-wide RBAC to many resources, so a slim and secure (implied by slim) is preferred. Ideally without a shell. In other projects I work on, we are using https://github.com/GoogleContainerTools/distroless as base, and to run smoothly on OpenShift clusters I would recommend the nonroot flavor.

takahiro-yamada:
@jwhitaker-gridcog 
Thank you for using this project and for your proposal!

Building cross-platform images sounds like a great idea, as it offers advantages with minimal risk or cost. We will consider it positively, so please give us a few days to discuss it with the core maintainers.

@erikgb 
Thank you for your continuous contributions! I personally support your suggestion to use a more secure, lightweight, and neutral base image. It's time to reconsider this, but could you give us some more time to investigate the impact of replacing the base image? I'd like to continue this discussion about the base image in #103.

takahiro-yamada:
@jwhitaker-gridcog 
We have discussed and decided to provide cross-platform images. While we can implement it, we would appreciate it if you could make a pull request to implement it as you suggested. How does that sound to you?

PR #156, opened 2025-03-27
Build multiarch image

jwhitaker-gridcog:
As per https://github.com/cybozu-go/accurate/issues/155 , this makes the CI build a multi-arch image.

The image is built using `buildx`, as per https://docs.docker.com/build/building/multi-platform/#prerequisites - Custom Builder .

On the dockerfile side, I've done it by using Go's cross-compilation abilities. The build stage is always host-native, and the binary is cross-compiled where necessary. This is the fastest way of doing things and works well for pure Go projects like this.

On the CI side, the Helm check needed to change a little bit: docker's local store can't handle multi-arch images, so instead of building to the local store and then using `kind load docker-image`, we set up a kind registry and push straight into that instead.

In addition, I've set the image to compress with `zstd` as it improves performance on container start a bit.

I can only test the release system so much as I don't have your ghcr keys :)

### DIFF
--- a/.github/workflows/helm.yaml
+++ b/.github/workflows/helm.yaml
@@ -18,6 +18,14 @@ jobs:
       - uses: actions/setup-python@v5
         with:
           python-version: 3.13
+
+      - name: set up a buildkit builder
+        # --driver docker-container is required to get a builder that can build multi-arch.
+        # --driver-opt network=host is required to be able to push to the kind registry, which is sitting on the host at :5000.
+        run: |
+          docker buildx create --name=b --driver docker-container --driver-opt network=host
+          docker buildx use b
+
       - name: Set up chart-testing
         uses: helm/chart-testing-action@0d28d3144d3a25ea2cc349d6e59901c4ff469b3b # v2.7.0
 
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -1,10 +1,14 @@
...
+# Build the manager binary. This always executes in the native architecture of the building machine.
+FROM --platform=$BUILDPLATFORM ghcr.io/cybozu/golang:1.23-jammy AS builder
 
 COPY ./ .
-RUN CGO_ENABLED=0 go build -ldflags="-w -s" -o accurate-controller ./cmd/accurate-controller
 
-# the controller image
+# Build the binary, cross-compiling if necessary
+ARG TARGETPLATFORM
+RUN CGO_ENABLED=0 GOOS=linux GOOARCH=$TARGETPLATFORM \
```
zoetrope, line 8:
```suggestion
RUN CGO_ENABLED=0 GOOS=linux GOARCH=$TARGETARCH \
```

`TARGETPLATFORM` contains the target os.
Please use `TARGETARCH` instead.

Dockerfile
```diff
@@ -1,10 +1,14 @@
...
-FROM ghcr.io/cybozu/golang:1.23-jammy as builder
+# Build the manager binary. This always executes in the native architecture of the building machine.
+FROM --platform=$BUILDPLATFORM ghcr.io/cybozu/golang:1.23-jammy AS builder
 
 COPY ./ .
-RUN CGO_ENABLED=0 go build -ldflags="-w -s" -o accurate-controller ./cmd/accurate-controller
 
-# the controller image
+# Build the binary, cross-compiling if necessary
+ARG TARGETPLATFORM
```
zoetrope, line 7:
```suggestion
ARG TARGETARCH
```

.github/workflows/helm.yaml
zoetrope, line 1: It seems like using a multi-architecture container image in this Helm test might not be necessary.
The existing test seems to work fine as it is.
If there's a particular reason for this changes, please let me know.

Dockerfile
zoetrope: When using TARGETPLATFORM, I got the following error.
```
go: unsupported GOOS/GOARCH pair linux/linux/amd64
```

jwhitaker-gridcog: oh goodness, yes this would not have worked. I did test the image but could only do so with docker's auto-binfmt thing, I'm guessing the x64 binary in the arm64 image just ran without emulation and I was none the wiser! Thanks for detailed checks on this :)

.github/workflows/helm.yaml
jwhitaker-gridcog: The purpose behind this was to ensure the test runs with as-close-as-possible-to the same image (and everything else) that is used in production, to try and maximize the chances of picking up if something breaks.

It's not perfect - it missed the broken arm64 subimage you found for example (it would have needed to be running on an arm runner to find this). However imo it's still worth testing a prod-like image if feasible.

zoetrope: I see. Thanks!

zoetrope (approved):
LGTM

PR #176, opened 2025-09-02
Fix Renovate authentication with GitHub App token

### LINKED ISSUE #166: Renovate Settings and Future Plan
## What  
- Renovate is currently not working due to a token authentication error.  
- Since we have few dependencies and do manual updates occasionally, there has been no major issue so far.  

## Discussion Points  
1. Do we want to keep using Renovate?  
2. If yes, is it worth fixing and maintaining?  
3. If no, what are the risks (e.g. missing security updates), and alternatives?

## Proposal  
Let’s decide based on the following flow:  
1. Decide whether to keep Renovate  
2. If yes:  
   - Investigate the token error and fix it  
3. If no:  
   - Remove or disable config files (`.github/renovate.json`, etc.)  
   - Set up a manual update process or consider alternatives  
4. Agree on the final direction 

## How  
- Check the cause of the token error  
- Clean up config files (delete or comment out)  
- Define and document an alternative update flow  

## Checklist  
- [ ] Decide whether to keep using Renovate  
- [ ] Investigate the token error  
- [ ] Remove or disable config files  
- [ ] Define an alternative update process  
- [ ] Update documentation (e.g. README)

012xx:
### Overview

- Resolved #166 
- Fix Renovate authentication by implementing GitHub App token-based setup as discussed in issue.

### What

- Implemented GitHub App token authentication for Renovate workflow
- Added `actions/create-github-app-token@v2` action to generate short-lived tokens
- Updated Renovate GitHub Action to use the generated app token
- Configured workflow to run on schedule (weekly on Monday at 10:00 UTC) and manual dispatch

### Why

Renovate was previously not working due to token authentication errors. 
We decided to continue using Renovate instead of switching to Dependabot because: 

- Renovate has more features compared to Dependabot
- Aqua dependencies can only be updated with Renovate
- GitHub App tokens provide better security than PATs (Personal Access Tokens) as they are short-lived and have limited scope
- This approach allows CI to be triggered on Renovate PRs, unlike using GITHUB_TOKEN

### Ref

When this CI is executed, the following dashboard is expected to be generated.
- #174

### DIFF
--- a/.github/workflows/renovate.yaml
+++ b/.github/workflows/renovate.yaml
@@ -18,7 +18,7 @@ jobs:
         with:
           token: ${{ steps.app-token.outputs.token }}
       - name: Self-hosted Renovate
-        uses: renovatebot/github-action@13da59cf7cfbd3bfea72ce26752ed22edf747ce9 # renovate v43.0.2
+        uses: renovatebot/github-action@7876d7a812254599d262d62b6b2c2706018258a2 # renovate v43.0.10
         with:
           configurationFile: renovate.json
           token: ${{ steps.app-token.outputs.token }}
--- a/.github/workflows/renovate.yaml
+++ b/.github/workflows/renovate.yaml
@@ -2,14 +2,23 @@ name: Renovate
 on:
   schedule:
     - cron: '0 10 * * 1'
+  workflow_dispatch:
 jobs:
   renovate:
     runs-on: ubuntu-latest
     steps:
+      - name: Generate GitHub App Token
+        id: app-token
+        uses: actions/create-github-app-token@v2
+        with:
+          app-id: ${{ secrets.RENOVATE_APP_ID }}
+          private-key: ${{ secrets.RENOVATE_PRIVATE_KEY }}
       - name: Checkout
         uses: actions/checkout@v4
+        with:
+          token: ${{ steps.app-token.outputs.token }}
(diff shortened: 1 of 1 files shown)


.github/workflows/renovate.yaml
```diff
@@ -2,14 +2,23 @@ name: Renovate
...
   schedule:
     - cron: '0 10 * * 1'
+  workflow_dispatch:
 jobs:
   renovate:
     runs-on: ubuntu-latest
     steps:
+      - name: Generate GitHub App Token
+        id: app-token
+        uses: actions/create-github-app-token@v2
```
erikgb, line 12: Nit: I think we should always use GHA digests. This is also part of https://docs.renovatebot.com/presets-config/#configbest-practices, which I think we should adopt as a preset, if not already doing so.

If we add this preset to the Renovate config, this improvement can be left for Renovate to suggest. I see we have other GHAs not pinned by digest.

erikgb (approved):

.github/workflows/renovate.yaml
012xx: @erikgb 
Thanks feedback! 
fix: [d76b4e5](https://github.com/cybozu-go/accurate/pull/176/commits/d76b4e59f38c00de623f239977d8d22f8cd12114)

PR #182, opened 2025-10-16
docs: Clean up markdown formatting and modify outdated descriptions

takahiro-yamada:
This PR updates documentations to fix markdown syntax and update outdated descriptions.

### DIFF
--- a/docs/overview.md
+++ b/docs/overview.md
@@ -55,7 +55,7 @@ Without propagating Role/RoleBinding, the tenant user would be able to do nothin
 - Inheriting labels and annotations creation/update from parent namespaces
 
     Namespace labels often play important roles.
-    For example, [Pod Security Admission](https://github.com/kubernetes/website/blob/dev-1.22/content/en/docs/concepts/security/pod-security-admission.md#pod-security-admission-labels-for-namespaces), a new feature planned for Kubernetes 1.22, uses Namespace labels to control security policies.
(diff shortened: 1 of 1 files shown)


docs/README.md
```diff
@@ -3,4 +3,4 @@
 Accurate is a Kubernetes controller for soft multi-tenancy environments.
 It is currently developed and maintained by [Cybozu](https://cybozu-global.com/).
 
-The repository is at https://github.com/cybozu-go/accurate .
+The repository is at <https://github.com/cybozu-go/accurate> .
```
012xx: ASK
I don't think this needs to be enclosed (I vaguely understand that it's probably better to enclose it, but I can't give a specific reason why), is it enclosed so that it can be mechanically recognized as a link?

docs/concepts.md
```diff
@@ -27,9 +27,10 @@ Mode is one of the following:
 
 ## Propagating generated resources (DEPRECATED)
 
-> [!WARNING]  
-> Propagating generated resources is a deprecated feature and is subject for
-> removal soon.
+<div class="warning">
```
zoetrope: ref. https://rust-lang.github.io/mdBook/format/mdbook.html#classwarning

docs/accurate-controller.md
```diff
@@ -53,7 +53,7 @@ watches:
 
 ## Command-line flags
 
-```
+```txt
```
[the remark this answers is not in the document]
takahiro-yamada: Sorry, I understand your point now. I've made them consistent.
Since it's a minor change, I rebased and force-pushed.

PR #204, opened 2026-03-11
Introduce golangci-lint and remove standalone staticcheck

### LINKED ISSUE #171: Add golangci-lint to CI
## What

I think we should add golangci-lint to our CI. I am personally a huge fan of linting, as it tends to improve the code quality. We could start easy with only the linters enabled in golangci-lint by default.

The reason I want use the golangci-lint lint runtime, is that it will allow us to add [KAL](https://github.com/kubernetes-sigs/kube-api-linter) as a follow-up.

## How

We need to create the initial project golangci-lint configuration and fix all violations in the code. Then we need to decide how to run golangci-lint in CI. There is an official golangci-lint action, https://github.com/golangci/golangci-lint-action, but I am unsure if this will allow us to add KAL to golangci-lint. The alternative would be to write our custom GH workflow for golangci-lint. For using KAL with golangci-lint, KAL was just added to Gateway API, so we could use https://github.com/kubernetes-sigs/gateway-api/blob/main/.github/workflows/kal.yml as a reference.

We should also make it easy for contributors to run golangci-lint, so we should probably also add a new Makefile `lint` target.

## Checklist

- [ ] Finish implementation of the issue
- [ ] Test all functions
- [ ] Have enough logs to trace activities
- [ ] Notify developers of necessary actions

ystkfujii:
This PR introduces golangci-lint to the repository.

Changes:

- Add golangci-lint configuration
- Fix existing lint issues
- Remove the standalone staticcheck step, since staticcheck is already included in golangci-lint's default linters

Fixes #171

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -1,0 +1,8 @@
+version: "2"
+linters:
+  settings:
+    errcheck:
+      exclude-functions:
+        - fmt.Fprintf
+        - fmt.Fprintln
+        - (*text/tabwriter.Writer).Flush
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,6 @@ CTRL_RUNTIME_VERSION := $(shell awk '/sigs.k8s.io\/controller-runtime/ {print su
 
 # Test tools
 BIN_DIR := $(shell pwd)/bin
-GOLANGCI_LINT := $(BIN_DIR)/golangci-lint
 SUDO = sudo
 
 # Set the shell used to bash for better error handling.
@@ -100,12 +99,12 @@ envtest: setup-envtest
 		go test -v -count 1 -race ./hooks/... -ginkgo.show-node-events -ginkgo.v
 
 .PHONY: lint
-lint: tools
-	$(GOLANGCI_LINT) run ./... -v
+lint: setup
+	golangci-lint run ./... -v
 
 .PHONY: lint-fix
-lint-fix: tools
-	$(GOLANGCI_LINT) run ./... -v --fix
+lint-fix: setup
+	golangci-lint run ./... -v --fix
 
 .PHONY: test
 test: 
@@ -148,9 +147,3 @@ GOBIN=$(PROJECT_DIR)/bin go install $(2) ;\
 }
 endef
 
-.PHONY: tools
-tools: $(GOLANGCI_LINT)
-
-$(GOLANGCI_LINT):
-	mkdir -p $(BIN_DIR)
-	GOBIN=$(BIN_DIR) go install github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.11.3
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,6 @@ CTRL_RUNTIME_VERSION := $(shell awk '/sigs.k8s.io\/controller-runtime/ {print su
 
 # Test tools
 BIN_DIR := $(shell pwd)/bin
-STATICCHECK := $(BIN_DIR)/staticcheck
 GOLANGCI_LINT := $(BIN_DIR)/golangci-lint
 SUDO = sudo
 
@@ -103,7 +102,6 @@ envtest: setup-envtest
 .PHONY: lint
 lint: tools
 	$(GOLANGCI_LINT) run ./... -v
-	$(STATICCHECK) ./...
 
 .PHONY: test
 test: 
@@ -147,11 +145,7 @@ GOBIN=$(PROJECT_DIR)/bin go install $(2) ;\
 endef
 
 .PHONY: tools
-tools: $(STATICCHECK) $(GOLANGCI_LINT)
-
-$(STATICCHECK):
-	mkdir -p $(BIN_DIR)
-	GOBIN=$(BIN_DIR) go install honnef.co/go/tools/cmd/staticcheck@latest
+tools: $(GOLANGCI_LINT)
 
 $(GOLANGCI_LINT):
 	mkdir -p $(BIN_DIR)
--- a/Makefile
+++ b/Makefile
@@ -103,6 +103,10 @@ envtest: setup-envtest
 lint: tools
 	$(GOLANGCI_LINT) run ./... -v
 
+.PHONY: lint-fix
+lint-fix: tools
+	$(GOLANGCI_LINT) run ./... -v --fix
+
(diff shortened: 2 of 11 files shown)


.golangci.yaml
```diff
@@ -0,0 +1,8 @@
+version: "2"
+linters:
+  settings:
+    errcheck:
+      exclude-functions:
```
ystkfujii, line 5: Note: `errcheck.exclude-functions` may look broad, but kubectl also does not handle write errors individually in CLI commands.

Example:
https://github.com/kubernetes/kubectl/blob/508904fd5941384440f72d0bc580b6a490bfc591/pkg/cmd/version/version.go#L145

Makefile
```diff
@@ -141,9 +144,9 @@ GOBIN=$(PROJECT_DIR)/bin go install $(2) ;\
...
-.PHONY: test-tools
-test-tools: $(STATICCHECK)
+.PHONY: tools
+tools: $(GOLANGCI_LINT)
 
-$(STATICCHECK):
+$(GOLANGCI_LINT):
 	mkdir -p $(BIN_DIR)
-	GOBIN=$(BIN_DIR) go install honnef.co/go/tools/cmd/staticcheck@latest
+	GOBIN=$(BIN_DIR) go install github.com/golangci/golangci-lint/v2/cmd/golangci-lint@latest
```
erikgb: I think we should lock the golangci-lint version. Breaking changes are not uncommon.

controllers/propagate.go
```diff
@@ -143,7 +143,7 @@ func (r *PropagateController) Reconcile(ctx context.Context, req ctrl.Request) (
 			return ctrl.Result{}, err
 		}
 	case "":
-		//lint:ignore SA1019 subject for removal
+		//nolint:staticcheck
```
erikgb: The comment about why it's suppressed should retain.

controllers/propagate.go
```diff
@@ -397,7 +397,7 @@ func (r *PropagateController) SetupWithManager(mgr ctrl.Manager) error {
 		if _, ok := ann[constants.AnnPropagate]; ok {
 			return true
 		}
-		//lint:ignore SA1019 subject for removal
+		//nolint:staticcheck
```
erikgb: Keep comment

erikgb (commented):
Good move! 👍 I think we should add a "lint-fix" make target. It's very handy for checks that support auto-fixing.

Makefile
ystkfujii: Fixed.

controllers/propagate.go
ystkfujii: Done

controllers/propagate.go
ystkfujii: done

ystkfujii:
@erikgb 

Thanks for the quick review! I've pushed updates to address the comments. 
Please take a look when you have time.

Makefile
```diff
@@ -141,9 +148,9 @@ GOBIN=$(PROJECT_DIR)/bin go install $(2) ;\
...
-.PHONY: test-tools
-test-tools: $(STATICCHECK)
+.PHONY: tools
+tools: $(GOLANGCI_LINT)
 
-$(STATICCHECK):
+$(GOLANGCI_LINT):
 	mkdir -p $(BIN_DIR)
-	GOBIN=$(BIN_DIR) go install honnef.co/go/tools/cmd/staticcheck@latest
+	GOBIN=$(BIN_DIR) go install github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.11.3
```
erikgb: Should the golangci-lint version be put in a variable that can be renovated by Renovate?

ystkfujii: @erikgb 

I agree. It should be managed by Renovate.
Which approach do you prefer?

- Remove the golangci-lint setting from the Makefile and manage it via aqua.yaml
- Keep the version in the Makefile and manage it with Renovate's regex manager

Personally, I think managing it via aqua.yaml would be simpler.

erikgb: I personally don't use or know aqua. But the version should be managed by Renovate. Please do what you think works best.

ystkfujii: Thanks for your comment.
I added a change ( https://github.com/cybozu-go/accurate/pull/204/changes/4ba3171ca8e40b831d12a8622027eb2c5f400942 ) to manage the golangci-lint version in aqua.yaml.
This version will also be managed by Renovate.

Just to be safe, I ran Renovate in local mode to verify it, and it detected the version correctly.
(For testing, I set it to v2.11.2, which is slightly older than the latest version.)

<details>

```
$ docker run --rm -e GITHUB_COM_TOKEN=${GITHUB_COM_TOKEN}  -v "$PWD:/work"   -w /work   -e LOG_LEVEL=debug   renovate/renovate:latest   --platform=local 
...
...
               {
                 "depName": "golangci/golangci-lint",
                 "packageName": "golangci/golangci-lint",
                 "currentValue": "v2.11.2",
                 "datasource": "github-tags",
                 "replaceString": " name: golangci/golangci-lint@v2.11.2",
                 "updates": [
                   {
                     "bucket": "non-major",
                     "newVersion": "v2.11.3",
                     "newValue": "v2.11.3",
                     "releaseTimestamp": "2026-03-10T10:21:26.000Z",
                     "newVersionAgeInDays": 1,
                     "newMajor": 2,
                     "newMinor": 11,
                     "newPatch": 3,
                     "updateType": "patch",
                     "isBreaking": false,
                     "pendingChecks": true,
                     "libYears": 0.006960616438356164,
                     "branchName": "renovate/all-minor-patch"
                   }
                 ],
                 "versioning": "semver-coerced",
                 "warnings": [],
                 "sourceUrl": "https://github.com/golangci/golangci-lint",
                 "registryUrl": "https://github.com",
                 "mostRecentTimestamp": "2026-03-10T10:21:26.000Z",
                 "isAbandoned": false,
                 "currentVersion": "v2.11.2",
                 "currentVersionTimestamp": "2026-03-07T21:22:56.000Z",
                 "currentVersionAgeInDays": 3,
                 "isSingleVersion": true,
                 "fixedVersion": "v2.11.2"
               }
...
...
```

</details>

erikgb: Nice! Thanks for testing!

erikgb (approved):
LGTM

Issue #103, opened 2023-10-26
Upgrade/patch/replace Go build image


erikgb:
## What

The image used to build the operator binary seems to be a Cybozu Ubuntu Jammy based image:

https://github.com/cybozu-go/accurate/blob/ad5cd954edf6d1bddde64df666448ac1c3ad5661/Dockerfile#L2

Is there any particular reason for not using the official Go image, https://hub.docker.com/_/golang?

## How

Replace the Go binary build image with `golang:1.21`, eventually an exact image digest if we want reproducible builds.

## Checklist

- [ ] Finish implementation of the issue
- [ ] Test all functions
- [ ] Have enough logs to trace activities
- [ ] Notify developers of necessary actions

zoetrope:
I've upgraded to go1.21 for now. #105

While I'm using `quay.io/cybozu/golang` in accordance with our company policy, I believe we could also use the official golang if needed.

erikgb:
> While I'm using `quay.io/cybozu/golang` in accordance with our company policy, I believe we could also use the official golang if needed.

We would prefer a more open/neutral base image. Is this base image open-sourced?

yamatcha:
Base image is built in this open-source repository.
https://github.com/cybozu/neco-containers/tree/main/golang-all

takahiro-yamada:
@erikgb 
As you mentioned in comment on #155, we have concluded that it's better to use a more lightweight and secure image as the base image. Before we start the implementation, we would like to confirm a few points.

The `cybozu-go/accurate` currently uses the `cybozu/golang` image as the builder base and scratch as the controller base.
https://github.com/cybozu-go/accurate/blob/aebac98b2a1d066dc4bbaf93ddd8e358c9ddd37a/Dockerfile#L1-L8

Which base image do you suggest we replace, and with what image, or should we replace both? While scratch is the most secure base, would you prefer to use another image with more utilities, such as `distroless`?

erikgb:
Thanks for you feedback, @takahiro-yamada! I didn't notice that we build the runtime image from scratch. That's fine with me.  👍 For the build image, I think anything that works is good. Thanks for clarifying this and explaining how this is set up. I'll just close the issue now. Feel free reopen if you want to change anything.

PR #208, opened 2026-03-27
Bump dependencies version

012xx:
### Overview

Bump dependencies version

### What Updates

- `actions/*`
- `aquaproj/aqua-registry`
- `mikefarah/yq`
- `helm/helm`
- `bitnami-labs/sealed-secrets`
- `goreleaser/goreleaser`
- `golangci/golangci-lint`
- `go.mod`


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/cybozu-go/accurate
 
-go 1.25.0
+go 1.25.5
```
rkwmt: How about using the latest 1.25.8? As far as I tested locally, it built successfully.

rkwmt (changes requested):
Since there is a newer version 4.0.4, how about using this one?
https://github.com/aquaproj/aqua-installer/releases/tag/v4.0.4
https://github.com/cybozu-go/accurate/blob/3b9c029870829747aa1d2cc83620270c2d30a15f/.github/actions/aqua/action.yaml#L10

go.mod
012xx: Thanks! fixed [afa2814](https://github.com/cybozu-go/accurate/pull/208/commits/afa2814bf9779b5ffaee87d982a4ef98d7f1cec1)

erikgb: It is not common to bump the **minimum** Go version when not required to. I suggest we keep this at 1.25.0.

ystkfujii: @erikgb 
I just want to make sure I understand your intent correctly.
Are you suggesting that we should specify the minimum supported Go version?
For example, does that mean we should keep it at 1.25 until Go 1.27 is released?
Also, is the idea of keeping the patch version at .0 for the same reason as [why not bump on each minor release](https://go.googlesource.com/proposal/+/HEAD/design/69095-x-repo-continuous-go.md#why-not-bump-on-each-minor-release) ?

erikgb: I think we should leave this as the minimum Go version required to **build** Accurate. Even if a Go version goes out of support, we should stick to the minimum version required to build the Go module, but this will never happen to Accurate, as we depend upon Kubernetes Go modules that usually require newer Go features. But for simpler Go modules, this can be left far behind. As an example, take a look at https://github.com/stretchr/testify/blob/5f80e4aef7bee125b7e9c0b620edf25f6fc93350/go.mod#L3-L5 as an example.

But we should, of course, build official releases of Accurate with the latest Go available.

See also the official Go documentation: https://go.dev/ref/mod#go-mod-file-go

ystkfujii: @erikgb 

Sorry for the late reply.

I understand the reasoning behind keeping the `go` directive version as low as possible. 
I also agree that official releases should be built with the latest available Go version.

In that case, I think we should specify the Go version used for builds separately from the `go` directive. We could do that either in `setup-go` or with the `toolchain` directive, but I’m leaning toward using `toolchain` so that everything stays in `go.mod`.

```go
go 1.25.0
toolchain go1.25.8
```

The approach I have in mind is to update `toolchain` proactively, while only bumping `go` to match the requirements of the modules we depend on.

To be honest, I’m still not sure how far back we should go in terms of the minimum supported Go version. That said, since we depend on Kubernetes modules, I do not expect us to support very old Go versions in practice, so I think this approach should work fine.

What do you think about this approach?

erikgb: I would personally prefer to avoid using the `toolchain` at all. I've had a lot of issues with it, and think it's not well-designed. Why are you reluctant to control the actual Go version used to build the image outside of go.mod?

The actual Go version used to build is now controlled by the closed-source image referenced [here](https://github.com/cybozu-go/accurate/blob/38910f43ba7d5f3603e1994fe65cc65bbaa40868/Dockerfile#L2), which I never liked, ref.  https://github.com/cybozu-go/accurate/issues/103. If we switched this image to the public Go image, it should be easy to ensure we always build with the latest version of Go.

ystkfujii: I see, that does make things a little tricky... 😅

> Why are you reluctant to control the actual Go version used to build the image outside of go.mod?

I think it would be more natural to follow the intended roles of `go` and `toolchain` here.

[the Go toolchain documentation](https://go.dev/doc/toolchain#config) describes it as follows:

> The go line declares the minimum required Go version for using the module or workspace. 
> ...
> The toolchain line declares a suggested toolchain to use with the module or workspace. 
> ...
> toolchain line is omitted, the module or workspace is considered to have an implicit toolchain goV line, where V is the Go version from the go line.

So if we keep only the minimum supported version in go, Accurate also ends up implicitly recommending that same lower version. That seems a bit different from the intention, where the minimum supported version stays low while builds and releases use a newer Go version.

There is no concrete issue today, but I think this kind of mismatch could become confusing for tools that read `go.mod`. For example, some tools such as `osv-scanner` interpret the Go standard library version from `go.mod`. 

REF: https://github.com/google/osv-scalibr/blob/39707fc6c55a96505ec166dc27520a56b6fb17a6/extractor/filesystem/language/golang/gomod/gomod.go#L186-L205
 
Even if we do not use such tools directly in Accurate, I think it is better to stay aligned with how Go expects these fields to be used.Because of that, I think the version Accurate recommends should be stated in `go.mod`.

To me, there are two natural options:

1. Do not use `toolchain`, and update `go` itself to the latest version.
2. Keep `go` as the minimum required version, and use `toolchain` for the latest version.

Personally, I would be fine with either approach.

> If we switched this image to the public Go image, it should be easy to ensure we always build with the latest version of Go.

As for the closed-source image, there are currently internal policy constraints, so changing it while the project is still under the cybozu-go org would not be easy. If the project is donated to CNCF in the future, I think that part could be revisited then.

erikgb: What has annoyed me most with the Go toolchain feature is summarized in [the docs](https://go.dev/doc/toolchain):

> To run a toolchain other than the bundled Go toolchain, the go command searches the process’s executable path ($PATH on Unix and Plan 9, %PATH% on Windows) for a program with the given name (for example, go1.21.3) and runs that program. If no such program is found, the go command [downloads and runs the specified Go toolchain](https://go.dev/doc/toolchain#download). Using the GOTOOLCHAIN form <name>+path disables the download fallback, causing the go command to stop after searching the executable path.

Attempting to download Go in CI is just wrong IMO. 😓 

> As for the closed-source image, there are currently internal policy constraints, so changing it while the project is still under the cybozu-go org would not be easy. If the project is donated to CNCF in the future, I think that part could be revisited then.

The proprietary image has been problematic on several occasions, as it seems to be lagging on Go versions. But we can postpone this discussion. 😄 Building with the latest version of Go is most important to me.

ystkfujii: I agree that we should avoid any behavior that automatically downloads Go in CI. 👍
That said, I think this is a slightly separate issue from whether we use the toolchain directive in go.mod.

As I understand it, even if toolchain is set in go.mod, Go will not be downloaded as long as `GOTOOLCHAIN=local` is used. In fact, in our GitHub Actions workflow, we determine the Go version from go.mod and run with `GOTOOLCHAIN=local`. In that setup, automatic toolchain download does not happen, and we have already seen cases where execution failed for that reason.

https://github.com/cybozu-go/accurate/actions/runs/23952642132/job/69863562341#step:8:11

On the other hand, Docker is probably still running with the default `GOTOOLCHAIN=auto`, so if it is executed with a Go version lower than the one specified by toolchain, it may try to download Go there. But even that should be avoidable by explicitly setting `GOTOOLCHAIN=local`.


<details>
<summary>When the local Go version is lower than the go directive in go.mod.</summary>


```
$ docker run -it -v $(pwd):/wd -w /wd --rm ghcr.io/cybozu/golang:1.24-jammy bash
root@b259fcfa0463:/wd# head -5 go.mod 
module github.com/cybozu-go/accurate

go 1.25.0

toolchain go1.26.0
root@b259fcfa0463:/wd# 
root@b259fcfa0463:/wd# #With GOTOOLCHAIN=local, it results in an error without attempting to download go
root@b259fcfa0463:/wd# GOTOOLCHAIN=local go build -o out cmd/kubectl-accurate/main.go 
go: go.mod requires go >= 1.25.0 (running go 1.24.13; GOTOOLCHAIN=local)
root@b259fcfa0463:/wd# echo $?
1
root@b259fcfa0463:/wd# # With the default setting (GOTOOLCHAIN=auto), a download occurs.
root@b259fcfa0463:/wd# go build -o out cmd/kubectl-accurate/main.go 
go: downloading go1.26.0 (linux/amd64)
go: downloading github.com/spf13/cobra v1.10.2
...
root@b259fcfa0463:/wd#
```

</details>


<details>
<summary>When the local Go version is higher than the go directive in go.mod.</summary>

```
$ docker run -it -v $(pwd):/wd -w /wd --rm ghcr.io/cybozu/golang:1.25.8.1_jammy bash
root@2f8dd2980b23:/wd# head -5 go.mod 
module github.com/cybozu-go/accurate

go 1.25.0

toolchain go1.26.0
root@2f8dd2980b23:/wd#
root@2f8dd2980b23:/wd# # With GOTOOLCHAIN=local, no download occurs.
root@2f8dd2980b23:/wd# GOTOOLCHAIN=local go build -o out cmd/kubectl-accurate/main.go 
go: downloading github.com/spf13/cobra v1.10.2
...
root@2f8dd2980b23:/wd# 
root@2f8dd2980b23:/wd# # With the default setting (GOTOOLCHAIN=auto), a download occurs.
root@2f8dd2980b23:/wd# go build -o out cmd/kubectl-accurate/main.go 
go: downloading go1.26.0 (linux/amd64)
root@2f8dd2980b23:/wd# 
```

</details>

So at least to me, these are two separate questions:

- whether to use toolchain in go.mod
- whether to allow automatic Go downloads in CI or Docker

If there are no other concerns, I would like to move forward the approach I proposed earlier:

> The approach I have in mind is to update toolchain proactively, while only bumping go to match the requirements of the modules we depend on.

erikgb: Choosing the lesser evil, I would vote in favour of bumping the minimum Go version (possibly prematurely) over introducing the Go toolchain directive.

ystkfujii: Thanks, I appreciate the clarification. I think that sounds like a good approach as well 👍 
I may still be missing part of the picture, so if anything turns out to have been overlooked, I’d be happy to continue the discussion in a separate issue or PR 

@012xx 

Based on the discussion above, it seems we should stick with the original plan, not introduce `toolchain`, and continue updating the `go` directive.

Also, to avoid any unnecessary downloads during the build, please keep the version in the `go` directive aligned with the Dockerfile base image version.

Could you take care of that change as well?

erikgb: Thanks @ystkfujii! Appreciated!

ystkfujii: FYI, I suggested setting `GOTOOLCHAIN=local` on the Go image side:
- https://github.com/cybozu/neco-containers/issues/1896

012xx: fixed [cf929e0](https://github.com/cybozu-go/accurate/pull/208/commits/cf929e0cebc935427bff10eea7a432a2c4d95176)

rkwmt (approved):

Issue #173, opened 2025-08-09
New major (cleanup) release of Accurate


erikgb:
I want to propose a new major (breaking) release of Accurate to clean up and improve various things. We have both features and and API versions that has been deprecated long enough to justify removal.

Here is a detailed list of things we should clean up (or improve):

- [ ] Remove code to upgrade managed fields from CSA to SSA.
- [ ] Remove deprecated SubNamespace API versions (v1 and v2alpha1).
- [ ] Remove the deprecated `installCRDs` value in our Helm chart.
- [ ] Remove the deprecated "propagate-generated" feature.
- [ ] Only publish our Helm chart as an OCI artifact, ref. https://github.com/cybozu-go/accurate/issues/157.
- [ ] Remove the client-side rate-limiting options, ref. https://github.com/cybozu-go/accurate/issues/165.

ystkfujii:
Is there anything else we need to do before making the new major release?

If not, I plan to create the `v2.0.0` tag at any time after https://github.com/cybozu-go/accurate/pull/235 is merged.

erikgb:
> Is there anything else we need to do before making the new major release?

I've just taken another look, and this looks great. A few things to consider:

1. Should we migrate the project to kubebuilder [go/v4](https://book-v3.book.kubebuilder.io/migration/migration_guide_gov3_to_gov4#migration-from-gov3-to-gov4) layout? Another option to consider is to make a dedicated API Go module. This is becoming more common in operator projects, and a huge benefit if another operator is importing our APIs. I would personally vote in favour of adopting as much as possible from the kubebuilder go/v4 layout, but with a separate Go module for our API.
2. What do you think about adopting [release-please](https://github.com/googleapis/release-please)? This will definitely streamline our release process and make it easier to cut new releases as needed.

ystkfujii:
@erikgb 

Sorry for the late reply, and thanks for the suggestions 🙇 

For 1, adopting as much as possible from the Kubebuilder go/v4 layout sounds good to me. 
However, I am not yet sure whether we should also split the API into a separate Go module.

Just to make sure I understand correctly, do you mean adding a separate `go.mod` under `api/`, using a module path such as `github.com/cybozu-go/accurate/api`, and having the main module depend on it?
Is the main benefit that other operators could import our API types without depending on the rest of Accurate and its unrelated dependencies?

For 2, I tried release-please a little, and it looks useful. I am in favor of adopting it.

[...]

PR #127, opened 2024-05-15
fix: upgrade SubNamespace status managed fields from CSA to SSA

erikgb:
This PR fixes the leftovers from https://github.com/cybozu-go/accurate/pull/121 and completes the operator migration from CSA to SSA. We should at some point remove the CSA to SSA upgrade code - as noted in the migration TODOs.

Would it be possible to cut a new release soon?

PR #210, opened 2026-04-03
Remove obsolete SSA migration code

erikgb:
We migrated this operator to use server-side apply (SSA) a long time ago, so I think it's time to remove the code related to upgrading from CSA to SSA.

Relates to https://github.com/cybozu-go/accurate/issues/173.


takahiro-yamada (approved):
LGTM, thank you!

---
This PR reverts some parts introduced in #112, #121, #127. The migration period is over, so this cleanup looks good to me.

PR #213, opened 2026-04-06
Manage setup-envtest version in aqua.yaml

ystkfujii:
## Summary

- Add `setup-envtest` (v0.23.3) to `aqua.yaml` to pin the version explicitly
- Remove the `setup-envtest` Make target that previously installed it via `go install @latest`
- Enable aqua checksum verification and add `aqua-checksums.json`

## Note

CI is currently failing but will be fixed by this change.
- https://github.com/cybozu-go/accurate/actions/runs/23952642132/job/69863562341

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -90,7 +90,7 @@ check-generate:
 	git diff --exit-code --name-only
 
 .PHONY: envtest
-envtest: 
+envtest: setup
 	source <(setup-envtest use -p env); \
 		TEST_CONFIG=1 go test -v -count 1 -race ./pkg/config -ginkgo.show-node-events -ginkgo.v -ginkgo.fail-fast
 	source <(setup-envtest use -p env); \
--- a/Makefile
+++ b/Makefile
@@ -90,12 +90,12 @@ check-generate:
 	git diff --exit-code --name-only
 
 .PHONY: envtest
-envtest: setup-envtest
-	source <($(SETUP_ENVTEST) use -p env); \
+envtest: 
+	source <(setup-envtest use -p env); \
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -129,21 +129,3 @@ release-build: setup
...
-SETUP_ENVTEST := $(shell pwd)/bin/setup-envtest
-.PHONY: setup-envtest
-setup-envtest: $(SETUP_ENVTEST) ## Download setup-envtest locally if necessary
-$(SETUP_ENVTEST):
-	# see https://github.com/kubernetes-sigs/controller-runtime/tree/master/tools/setup-envtest
-	GOBIN=$(shell pwd)/bin go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest
-
-# go-get-tool will 'go get' any package $2 and install it to $1.
-PROJECT_DIR := $(shell dirname $(abspath $(lastword $(MAKEFILE_LIST))))
-define go-get-tool
```
ystkfujii, line 142: It doesn’t seem to be used. Also, if we’re going to use `go install`, it would probably be better to use go tool or aqua instead

Makefile
```diff
@@ -90,12 +90,12 @@ check-generate:
 	git diff --exit-code --name-only
 
 .PHONY: envtest
-envtest: setup-envtest
-	source <($(SETUP_ENVTEST) use -p env); \
+envtest:
```
012xx: When I tried it locally, I got a setup-envtest: command not found error, so I thought envtest: setup might be necessary.

ystkfujii:
```suggestion
envtest: setup
```

ystkfujii: done

012xx (approved):
LGTM🚀

PR #225, opened 2026-06-11
Update Renovate workflow to use client ID

ystkfujii:
Client ID is recommended instead of App ID.

Ref: https://github.com/actions/create-github-app-token/pull/353

### DIFF
--- a/.github/workflows/renovate.yaml
+++ b/.github/workflows/renovate.yaml
@@ -9,7 +9,7 @@ jobs:
     steps:
       - name: Generate GitHub App Token
         id: app-token
-        uses: actions/create-github-app-token@f8d387b68d61c58ab83c6c016672934102569859 # v3.0.0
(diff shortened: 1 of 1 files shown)


.github/workflows/renovate.yaml
```diff
@@ -11,7 +11,7 @@ jobs:
         id: app-token
         uses: actions/create-github-app-token@f8d387b68d61c58ab83c6c016672934102569859 # v3.0.0
```
rkwmt: It looks like the `client-id` input was added in [v3.1.0](https://github.com/actions/create-github-app-token/releases/tag/v3.1.0). Wouldn't it be necessary to bump the version accordingly?

rkwmt (changes requested):

.github/workflows/renovate.yaml
ystkfujii: Oh... Its too old ...

ystkfujii: i fixed it

rkwmt (approved):
LGTM

PR #226, opened 2026-06-12
[CI] Add pinact check for GitHub Actions

ystkfujii:
## Overview:

This PR adds a CI workflow to verify pinned GitHub Actions with pinact.

The workflow checks that GitHub Actions SHAs and their version comments are consistent. 
It also adds pinact to aqua.

### DIFF
--- a/.github/workflows/pinact.yaml
+++ b/.github/workflows/pinact.yaml
@@ -1,0 +1,20 @@
+name: Pinact
+on:
+  pull_request:
+    paths:
+      - '.github/**/*.yml'
+      - '.github/**/*.yaml'
+jobs:
+  verify-pinact:
+    name: Verify pinned GitHub Actions comments
+    runs-on: ubuntu-slim
+    permissions:
+      contents: read
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      - uses: ./.github/actions/aqua
+      - name: Verify SHAs and tag comments are consistent
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          pinact run --check --verify-comment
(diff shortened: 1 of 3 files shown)


012xx (changes requested):
IMO
Please include information about pinact in the Readme.
Like this: https://github.com/cybozu-go/website-operator#update-pinned-github-actions

ystkfujii:
@012xx 

Currently, GHA is updated by renovate.
Are you suggesting that it should be updated by pinact instead of renovate ?

012xx:
@ystkfujii 
It seems to be set up to update via Renovate, but just like with cattage, it doesn't appear to be working right now.
I think we won't need to include it in the README if we fix it and update it via Renovate, but I thought it might be better to leave it written there until that point is corrected.

012xx:
I thought the following Action should have been updated, but I assumed Renovate wasn't running because my own update was the latest one.
https://github.com/cybozu-go/accurate/blame/7f54f3a8b4d3360dca70124f56d841b7d20fed3e/.github/workflows/ci.yaml#L15

However, I notice that the line below is indeed being updated by Renovate.
https://github.com/cybozu-go/accurate/blame/7f54f3a8b4d3360dca70124f56d841b7d20fed3e/.github/workflows/ci.yaml#L68C13-L68C36

If you are still using renovate to update GitHub Actions, it might be a good idea to state in the README file that you are using Renovate for dependency updates.

However, I think this is a matter of preference, so I would like to hear your opinion. If you think it should not be mentioned, I would like to go ahead and merge this as is.

ystkfujii:
@012xx 

Since it’s already documented in [maintenance.md](https://github.com/cybozu-go/accurate/blob/7f54f3a8b4d3360dca70124f56d841b7d20fed3e/docs/maintenance.md), it’s probably not needed in the README.

012xx (approved):
I missed that. In that case, it looks good.
LGTM🚀

PR #228, opened 2026-06-14
Remove long-term unserved and deprecated API versions

erikgb:
This PR was motivated by https://github.com/cybozu-go/accurate/pull/227. The deprecated API versions have been unserved for a long time, and I think it's time to remove them.

I also did some minor, unrelated changes:
- remove uneeded fix-crds patch
- updated API doc generation to use v2 instead of v1

This cleanup is mentioned in https://github.com/cybozu-go/accurate/issues/173.


ystkfujii:
Thank you for the suggestion. 
I agree that it would be a good time to remove them.

I’m planning to take care of this after https://github.com/cybozu-go/accurate/pull/227 is merged.

erikgb:
> I’m planning to take care of this after #227 is merged.

Why **after**, and not before? Deleting this code will make #227 a lot smaller and easier to review. 😅

config/crd-only/kustomization.yaml
```diff
@@ -3,5 +3,5 @@ labels:
       app.kubernetes.io/name: accurate
 resources:
 - ../crd
-patches:
-  - path: crd_conversion_patch.yaml
+#patches:
+#  - path: crd_conversion_patch.yaml
```
ystkfujii: [Q]

Do we still have a reason to keep it ?
If we are removing the conversion, it may no longer be necessary.

config/crd/kustomization.yaml
```diff
@@ -5,10 +5,9 @@ resources:
 - bases/accurate.cybozu.com_subnamespaces.yaml
 #+kubebuilder:scaffold:crdkustomizeresource
 
-patches:
-- path: patches/cainjection_in_subnamespaces.yaml
-- path: patches/fix-crd.yaml
-- path: patches/webhook_in_subnamespaces.yaml
+#patches:
+#- path: patches/cainjection_in_subnamespaces.yaml
+#- path: patches/webhook_in_subnamespaces.yaml
```
ystkfujii: Ditto

https://github.com/cybozu-go/accurate/pull/228/changes#r3480346285

config/kustomize-to-helm/overlays/crds/kustomization.yaml
```diff
@@ -2,7 +2,7 @@ resources:
   - ../../../crd
 
 patches:
-  - path: crd_conversion_patch.yaml
+#  - path: crd_conversion_patch.yaml
```
ystkfujii: Ditto

https://github.com/cybozu-go/accurate/pull/228/changes#r3480346285

config/crd/kustomization.yaml
ystkfujii: Ditto

https://github.com/cybozu-go/accurate/pull/228/changes#r3480346285

config/crd-only/kustomization.yaml
erikgb: I was trying to bring things back to how it's scaffolded by kubebuilder, but if you want, I can remove it. We might need a v3 in the future. 😉

ystkfujii: Hmm, this is a bit of a trade-off, but personally I think it would be better to remove it.

My main concerns are:
- Having the conversion manifest remain while the corresponding code no longer exists could be confusing.
- If we need it again in the future, we can add it back then.

That said, this is not a strong change request from my side, so I’m happy to leave the final decision to you. 
I’ll approve this PR.

ystkfujii (dismissed):

config/crd-only/kustomization.yaml
erikgb: I don't have strong opinions either, and can clean up the obsolete files.

ystkfujii: Okay, let’s remove the obsolete files. Thank you for cleaning this up!

erikgb: Done!

ystkfujii (approved):
LGTM

PR #236, opened 2026-07-10
Adopt Kubebuilder go/v4 layout (partially)

erikgb:
As discussed in https://github.com/cybozu-go/accurate/issues/173, this PR contains a (partial) adoption of the Kubebuilder go/v4 layout. I've tried to minimize the changes, so there is more to do as a follow-up.

It contains a possibly controversial change: I am proposing to drop the "constants" package and move some of the user-facing constants into the API package. This makes a lot of sense to me, but probably not for everyone.

I would have preferred to move/rename the pkg/config package, but I wasn't sure where/how. Any input would be highly welcome.

PS: I want to rebase and squash all commits when the form of this PR is agreed.


ystkfujii:
> I am proposing to drop the "constants" package and move some of the user-facing constants into the API package. 

I personally think it makes sense to keep the metadata under the `v2` directory.

At this point, it is not clear whether the metadata is specific to `v2`, so some people may feel that placing it under `v2` is not ideal. 
In that case, we could create a separate directory such as `metadata`.

However, I think that could become confusing if `v3`-specific metadata is added in the future.

Personally, I prefer this:

* `api/accurate/v2/meta.go`
* `api/accurate/v3/meta.go`

over this:

* `api/accurate/metadata/meta.go`
* `api/accurate/v2/meta.go`
* `api/accurate/v3/meta.go`

Even if some metadata is shared between versions, I think it is acceptable to define it separately for each version. We can decide how to share or reference it when `v3` is actually introduced.

> I would have preferred to move/rename the pkg/config package, but I wasn't sure where/how. Any input would be highly welcome.

I do not have a strong opinion about where this package should live.

If it is only used internally, moving it under `internal` makes sense. In that case, `internal/config/config.go` seems reasonable.
For `feature_gate.go`, it may also make sense to create a separate package such as `internal/feature`.

PROJECT
```diff
@@ -1,6 +1,7 @@
 domain: cybozu.com
 layout:
-- go.kubebuilder.io/v3
+- go.kubebuilder.io/v4
+multigroup: true
```
ystkfujii, line 4: [q]

Why do we need to enable `multigroup`? 🤔 
The directory structure in this PR looks different from the multi-group layout expected by Kubebuilder


https://book.kubebuilder.io/migration/multi-group#understanding-the-layouts

internal/indexing/constants.go
```diff
@@ -1,9 +1,10 @@
-package constants
+package indexing
 
 // Keys for indexing objects
 const (
 	NamespaceParentKey   = "namespace.parent"
 	NamespaceTemplateKey = "namespace.template"
 	PropagateKey         = "resource.propagate"
 	SubNamespaceNameKey  = "subnamespace.name"
+	PropagateAny         = "any" // defined as an in-memory index value
```
ystkfujii, line 9: [q]

Why was only this moved?

ystkfujii:
Hi @erikgb, I just wanted to check in on the status of this PR.

Do you have any plans to continue working on it? 
If not, I'd be happy to help move it forward on my side.

erikgb:
> Hi @erikgb, I just wanted to check in on the status of this PR.
> 
> Do you have any plans to continue working on it? If not, I'd be happy to help move it forward on my side.

@ystkfujii, thanks for the ping! To be honest, I have forgotten about this PR. 😅 Please feel free to pick it up, as I've been busy these days.

ystkfujii:
Addressed in https://github.com/cybozu-go/accurate/pull/239

Issue #237, opened 2026-07-29
ResourceQuota support


mhkarimi1383:
## What

It will be awesome if we could have support for Resource Qouta per Root Namespace

## How

We previously had the same in original project from K8s (retired)

## Checklist

- [ ] Finish implementation of the issue
- [ ] Test all functions
- [ ] Have enough logs to trace activities
- [ ] Notify developers of necessary actions

rkwmt:
@mhkarimi1383 
Thanks for the feature request!

We've actually discussed a few possible approaches for ResourceQuota support in the past, but there are several things we need to think through carefully, so we haven't implemented it yet.

We'll keep thinking about this, and we'd welcome any good ideas on how to approach it.

mhkarimi1383:
Hi

I have put the design that is in my mind in https://github.com/cybozu-go/accurate/pull/238

Thanks

PR #238, opened 2026-07-30
:memo: Initial DesignDoc for Resource Quota

mhkarimi1383:
This is the initial design document for the Resource Quota feature.

We need this to be able to discuss the design and implementation of the feature. This should be a good place to start.

Also in our Projects we need this feature to be able to use Accurate in Production.

> Note that structure of the document is made by AI, But the Content itself is from what was in my mind for designing this feature and How it should work

### DIFF
--- a/docs/quota.md
+++ b/docs/quota.md
@@ -1,0 +1,407 @@
+# Design Doc: Hierarchical Resource Quota for Accurate
+
+**Status:** Draft · **Author:** @mhkarimi1383 · **Last updated:** 2026-07-30
+
+---
+
+## 1. Background & Motivation
+
+[Accurate](https://github.com/cybozu-go/accurate) is a Kubernetes controller for soft multi-tenancy that (1) propagates namespace-scoped resources between namespaces and (2) lets tenant users create/delete sub-namespaces. Namespace trees are encoded purely with labels — `accurate.cybozu.com/type=root` marks a root, `accurate.cybozu.com/parent=<name>` marks a sub-namespace — and the `SubNamespace` CRD (`accurate.cybozu.com/v2`) is the user-facing handle for creating children. Crucially, Accurate does **not** use the `SubNamespace` resource to discover parent/child relationships; it walks the `parent` label so trees can be restructured freely. 【turn1fetch0】【turn2fetch1】【turn5fetch0】
+
+Resource propagation is opt-in: a resource annotated `accurate.cybozu.com/propagate=create|update` is copied from parent to sub-namespaces, and copies are stamped with `accurate.cybozu.com/from=<parent>`. The reconcilers are `NamespaceReconciler`, `SubNamespaceReconciler`, and one `PropagateController` per watched GVK configured in `config.yaml` (`watches`, `labelKeys`, `annotationKeys`, …). 【turn1fetch1】【turn3fetch0】
+
+**The gap.** A native `ResourceQuota` can be propagated today, but propagation only *copies the same object* into each child. Each child then independently enforces the parent's `hard` values, so the sum of children can vastly exceed the parent's intent — the opposite of a budget. Native `ResourceQuota` is also strictly per-namespace: `kube-controller-manager`'s quota controller only accounts usage in the namespace where the quota lives, with no subtree roll-up. 【turn2search10】【turn2search11】
+
+We need a quota primitive that bounds the **aggregate allocation of a namespace *and its entire sub-namespace tree***, with explicit support for "no limit" nodes and arbitrary tree depth.
+
+## 2. Goals & Non-Goals
+
+**Goals**
+
+- Bound total resource allocation (CPU, memory, and optionally pods/PVCs/services/configmaps/etc.) of a namespace *plus all of its descendants*.
+- Allow any node in the tree to declare "no limit" while still contributing its real usage to every ancestor's aggregate.
+- Preserve Accurate's opt-in, label-driven tree model; work with `kubectl accurate sub create/move/cut/graft`.
+- Surface live aggregate usage and remaining capacity in a `status` field per node.
+- Handle tree mutations (reparenting, cascading delete, root conversion) consistently.
+
+**Non-Goals**
+
+- Replacing native `ResourceQuota` for non-hierarchical use cases.
+- Hard multi-tenancy isolation (network, node-level, sandboxed runtimes).
+- GPU/device quota accounting beyond what `ResourceQuota` already supports (we mirror its resource list semantics).
+- Cross-cluster quota federation.
+
+## 3. Use Cases
+
+The canonical tree from the request:
+
+```
+Namespace a  (10 cores, 5 GiB)              ← root, capped
+└── Namespace b  (5 cores, 2.5 GiB)         ← sub, capped (≤ a)
+    └── … continue the tree
+
+Namespace c  (no limit)                      ← root, uncapped
+├── Namespace d  (2.5 cores, 1 GiB)          ← sub, capped
+└── Namespace e  (no limit)                  ← sub, uncapped
+```
+
+Semantics this must satisfy:
+
+| Node | Own HRQ hard | Subtree aggregate cap (effective) |
+|------|--------------|-----------------------------------|
+| `a`  | cpu=10, mem=5Gi | min(10/5Gi) over {a, b, …} |
+| `b`  | cpu=5, mem=2.5Gi | min(5/2.5Gi, ancestor 10/5Gi) over {b, …} |
+| `c`  | none | ∅ (unlimited) over {c, d, e} |
+| `d`  | cpu=2.5, mem=1Gi | 2.5/1Gi over {d} |
+| `e`  | none | ∅ (unlimited); only ancestor is `c` (also unlimited) |
+
+A Pod scheduled into `d` is bounded by `d`'s 2.5/1Gi. A Pod into `e` is bounded by nothing (both `e` and `c` are unlimited). A Pod into `b` is bounded by `b`'s 5/2.5Gi **and** by `a`'s 10/5Gi — i.e., if `a`'s subtree is already at 8 cores, `b` can only admit 2 more cores even though its own HRQ still says 5.
+
+## 4. Proposed Design
+
+### 4.1 New CRD: `HierarchicalResourceQuota` (HRQ)
+
+Introduce a new namespace-scoped CRD in the `accurate.cybozu.com` API group. An HRQ placed in namespace N declares the hard ceiling for **N's entire subtree** (N itself + all transitive descendants). Presence of an HRQ is opt-in; namespaces without one are "no limit" nodes.
+
+```yaml
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  name: hierarchicalresourcequotas.accurate.cybozu.com
+spec:
+  group: accurate.cybozu.com
+  names:
+    kind: HierarchicalResourceQuota
+    listKind: HierarchicalResourceQuotaList
+    singular: hierarchicalresourcequota
+    plural: hierarchicalresourcequotas
+    shortNames: [hrq]
+  scope: Namespaced
+  versions:
+    - name: v2
+      served: true
+      storage: true
+      subresources:
+        status: {}
+      schema:
+        openAPIV3Schema:
+          type: object
+          x-kubernetes-preserve-unknown-fields: true   # mirror ResourceQuota corev1.ResourceList
+          properties:
+            spec:
+              type: object
+              x-kubernetes-preserve-unknown-fields: true
+              properties:
+                hard:
+                  type: object
+                  description: "Hard limits for the namespace subtree (corev1.ResourceList)."
+                  x-kubernetes-preserve-unknown-fields: true
+                scopes:
+                  type: array
+                  items: { type: string }
+                  description: "Mirrors corev1.ResourceQuotaScope; optional."
+                scopeSelector:
+                  type: object
+                  x-kubernetes-preserve-unknown-fields: true
+            status:
+              type: object
+              x-kubernetes-preserve-unknown-fields: true
+              properties:
+                hard:
+                  type: object
+                  x-kubernetes-preserve-unknown-fields: true
+                used:
+                  type: object
+                  x-kubernetes-preserve-unknown-fields: true
+                subtreeUsed:
+                  type: object
+                  description: "Aggregate usage across this namespace and all descendants."
+                  x-kubernetes-preserve-unknown-fields: true
+                remaining:
+                  type: object
+                  description: "hard − subtreeUsed, clamped at 0."
+                  x-kubernetes-preserve-unknown-fields: true
+                children:
+                  type: array
+                  description: "Direct sub-namespaces accounted in subtreeUsed."
+                  items: { type: string }
+                observedDescendants:
+                  type: array
+                  items: { type: string }
+                conditions:
+                  type: array
+                  items:
+                    type: object
+                    x-kubernetes-preserve-unknown-fields: true
+```
+
+We intentionally mirror `corev1.ResourceQuotaSpec` (`hard`, `scopes`, `scopeSelector`) so users fluent with native `ResourceQuota` transfer directly, and so we can reuse `k8s.io/api/core/v1.ResourceList` and the same resource-name grammar (`requests.cpu`, `limits.memory`, `pods`, `persistentvolumeclaims`, …). 【turn2search10】
+
+### 4.2 Semantics
+
+1. **Subtree aggregate.** `status.subtreeUsed[N] = used[N] + Σ subtreeUsed[child]` for every direct child `c` whose `accurate.cybozu.com/parent` label points at this namespace. A node with no HRQ still reports `status.subtreeUsed` (computed by the controller) so ancestors can roll it up, but exposes no `hard`.
+2. **Effective ceiling.** For a namespace N, the effective ceiling is `min(N.spec.hard, min over ancestors A of A.spec.hard)` for each resource key. Ancestors without an HRQ (or without that key) impose no constraint for that key.
+3. **"No limit".** A namespace with no HRQ object, or an HRQ with an empty `spec.hard`, is unlimited. It is still counted into every ancestor's `subtreeUsed`.
+4. **Relationship to native `ResourceQuota`.** Native RQs and HRQs coexist. To avoid double accounting, an HRQ-managed resource key in a namespace **should not** also be enforced by a native `ResourceQuota` for the same key; the controller annotates any native RQ it owns with `accurate.cybozu.com/owned-hrq: <name>` and will not touch others. Users may keep native RQs for keys *not* covered by any HRQ.
+5. **Propagation.** An HRQ is **not** a propagated resource. It is a policy object. Putting `accurate.cybozu.com/propagate` on it has no effect — this is enforced by the propagator ignoring the `HierarchicalResourceQuota` GVK (it is not listed in `config.yaml.watches`).
+
+### 4.3 Controller architecture
+
+Two new reconcilers join the existing three (`NamespaceReconciler`, `SubNamespaceReconciler`, `PropagateController`): 【turn1fetch0】
+
+- **`HierarchicalResourceQuotaReconciler`** — owns the HRQ object: recomputes `status.used` (via shared quota evaluators), `status.subtreeUsed`, `status.remaining`, and the `children`/`observedDescendants` lists. Triggers re-enqueue of the parent HRQ (and ancestors) when subtree membership changes.
+- **`HRQUsageReconciler`** (a `PropagateController`-style multi-GVK watcher) — watches the same GVKs the core `resourcequota` controller does (Pod, PVC, Service, ConfigMap, Secret, RC, RS, Deployment, StatefulSet, …, plus `ResourceQuota` itself) but only in namespaces that are part of an HRQ-managed tree. On any change it enqueues the HRQ in that namespace and walks up the `parent` chain.
+
+A small helper, `pkg/hrq/tree.go`, provides:
+
+- `Descendants(ctx, ns) ([]string, error)` — BFS over namespaces where `accurate.cybozu.com/parent` chains down from `ns`.
+- `Ancestors(ctx, ns) ([]string, error)` — walk `parent` label up to a root.
+- `EffectiveHard(ctx, ns) (corev1.ResourceList, error)` — `min` over `ns` and ancestors.
+
+Tree discovery reuses Accurate's existing label convention (`accurate.cybozu.com/parent`, `accurate.cybozu.com/type=root`) so `sub move`/`cut`/`graft` automatically re-shape accounting without any new CLI. 【turn2fetch1】【turn5fetch0】
+
+### 4.4 Enforcement: admission webhook
+
+Accurate's design notes state *"No webhooks for propagated resources"* — the rationale being (a) any namespace-scoped resource can be propagated, so a Pod webhook risks chicken-and-egg bootstrapping, and (b) Accurate should not surprise users who didn't ask for limits. 【turn1fetch0】 HRQ enforcement is a different concern: it is an **opt-in quota policy**, structurally equivalent to native `ResourceQuota` (which is itself an admission plugin). The webhook therefore:
+
+- Is registered with `namespaceSelector` matching `accurate.cybozu.com/hrq-managed=true` (a label the controller sets on every namespace that is, or descends from, an HRQ-bearing namespace). Clusters without HRQs pay zero webhook cost.
+- Uses `failurePolicy=Fail` only for the quota-scoped create/update verbs; `failurePolicy` is configurable.
+- Intercepts the same verbs/GVKs the core `resourcequota` admission plugin does (Pod/PVC/PV-claim/Service/ConfigMap/…).
+- On admission, computes the projected `subtreeUsed` for each ancestor HRQ *as if the request were admitted*, and rejects with a `Forbidden` error citing the first HRQ that would be breached and the offending resource key.
+
+This keeps the model precise: the webhook is the single source of truth for "would this create exceed the subtree cap?" The `status` fields are advisory/observability and may lag briefly under contention; the webhook is authoritative.
+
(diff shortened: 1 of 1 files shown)


docs/quota.md
```diff
@@ -0,0 +1,407 @@
...
+
+## 7. Enforcement details
+
+- **Webhook object:** `ValidatingWebhookConfiguration accurate-hrq-validator`, scoped via `namespaceSelector: accurate.cybozu.com/hrq-managed=true`. Rules mirror the core `resourcequota` admission plugin's resource list (Pod, PVC, Service, ConfigMap, Secret, RC, RS, Deployment, StatefulSet, DaemonSet, Job, CronJob, …) on `create`/`update` (not `delete` — quota releases happen asynchronously via the usage reconciler).
+- **Projection logic:** for the admitted object, compute its `corev1.ResourceList` contribution using the same evaluator registry, add to the namespace's current `subtreeUsed`, then for each ancestor HRQ (and the namespace's own) check `subtreeUsed + delta ≤ hard` per key. Reject on the first breach, naming the HRQ, the namespace, and the key.
+- **Race window:** between webhook accept and usage reconciler update, two near-simultaneous creates can both pass. This is identical to native `ResourceQuota`'s well-known best-effort race and is acceptable for soft multi-tenancy. `status` is advisory; the webhook is the gate.
```
ystkfujii, line 341: Is this actually identical to native ResourceQuota behavior?

Could you explain why you consider native ResourceQuota to be best-effort in the same sense?

docs/quota.md
```diff
@@ -0,0 +1,407 @@
...
+5. **Propagation.** An HRQ is **not** a propagated resource. It is a policy object. Putting `accurate.cybozu.com/propagate` on it has no effect — this is enforced by the propagator ignoring the `HierarchicalResourceQuota` GVK (it is not listed in `config.yaml.watches`).
+
+### 4.3 Controller architecture
+
+Two new reconcilers join the existing three (`NamespaceReconciler`, `SubNamespaceReconciler`, `PropagateController`): 【turn1fetch0】
+
+- **`HierarchicalResourceQuotaReconciler`** — owns the HRQ object: recomputes `status.used` (via shared quota evaluators), `status.subtreeUsed`, `status.remaining`, and the `children`/`observedDescendants` lists. Triggers re-enqueue of the parent HRQ (and ancestors) when subtree membership changes.
+- **`HRQUsageReconciler`** (a `PropagateController`-style multi-GVK watcher) — watches the same GVKs the core `resourcequota` controller does (Pod, PVC, Service, ConfigMap, Secret, RC, RS, Deployment, StatefulSet, …, plus `ResourceQuota` itself) but only in namespaces that are part of an HRQ-managed tree. On any change it enqueues the HRQ in that namespace and walks up the `parent` chain.
```
ystkfujii, line 153: This design seems to reimplement much of Kubernetes ResourceQuota.

- https://kubernetes.io/docs/concepts/policy/resource-quotas/

Could we instead use native ResourceQuota to track usage in each namespace?

docs/quota.md
```diff
@@ -0,0 +1,407 @@
...
+
+Resource propagation is opt-in: a resource annotated `accurate.cybozu.com/propagate=create|update` is copied from parent to sub-namespaces, and copies are stamped with `accurate.cybozu.com/from=<parent>`. The reconcilers are `NamespaceReconciler`, `SubNamespaceReconciler`, and one `PropagateController` per watched GVK configured in `config.yaml` (`watches`, `labelKeys`, `annotationKeys`, …). 【turn1fetch1】【turn3fetch0】
+
+**The gap.** A native `ResourceQuota` can be propagated today, but propagation only *copies the same object* into each child. Each child then independently enforces the parent's `hard` values, so the sum of children can vastly exceed the parent's intent — the opposite of a budget. Native `ResourceQuota` is also strictly per-namespace: `kube-controller-manager`'s quota controller only accounts usage in the namespace where the quota lives, with no subtree roll-up. 【turn2search10】【turn2search11】
+
+We need a quota primitive that bounds the **aggregate allocation of a namespace *and its entire sub-namespace tree***, with explicit support for "no limit" nodes and arbitrary tree depth.
```
ystkfujii, line 15: The original request in https://github.com/cybozu-go/accurate/issues/237 is for a ResourceQuota per root namespace:

However, this proposal allows quotas on any namespace in the hierarchy.

Could you clarify the use case for setting quotas on intermediate sub-namespaces ?

---

Accurate allows tenant users to create and manage sub-namespaces. If platform administrators manage quotas on those namespaces, they need to follow a hierarchy controlled by tenant users.

If tenant users manage the quotas themselves, this becomes a quota delegation feature. In that case, the design should explain who can create, update, increase, or remove a child quota.

docs/quota.md
```diff
@@ -0,0 +1,407 @@
...
+
+- Is registered with `namespaceSelector` matching `accurate.cybozu.com/hrq-managed=true` (a label the controller sets on every namespace that is, or descends from, an HRQ-bearing namespace). Clusters without HRQs pay zero webhook cost.
+- Uses `failurePolicy=Fail` only for the quota-scoped create/update verbs; `failurePolicy` is configurable.
+- Intercepts the same verbs/GVKs the core `resourcequota` admission plugin does (Pod/PVC/PV-claim/Service/ConfigMap/…).
+- On admission, computes the projected `subtreeUsed` for each ancestor HRQ *as if the request were admitted*, and rejects with a `Forbidden` error citing the first HRQ that would be breached and the offending resource key.
+
+This keeps the model precise: the webhook is the single source of truth for "would this create exceed the subtree cap?" The `status` fields are advisory/observability and may lag briefly under contention; the webhook is authoritative.
```
ystkfujii, line 172: How do we keep the webhook state consistent across multiple Accurate replicas?

Reconcilers use leader election, but the webhook runs on every replica. Therefore, two webhook Pods could see different subtreeUsed values and both allow requests that exceed the quota.

docs/quota.md
```diff
@@ -0,0 +1,407 @@
...
+                  type: array
+                  items: { type: string }
+                conditions:
+                  type: array
+                  items:
+                    type: object
+                    x-kubernetes-preserve-unknown-fields: true
+```
+
+We intentionally mirror `corev1.ResourceQuotaSpec` (`hard`, `scopes`, `scopeSelector`) so users fluent with native `ResourceQuota` transfer directly, and so we can reuse `k8s.io/api/core/v1.ResourceList` and the same resource-name grammar (`requests.cpu`, `limits.memory`, `pods`, `persistentvolumeclaims`, …). 【turn2search10】
```
ystkfujii, line 138: I think the proposed public API is too broad for the initial implementation.

Could we start with a minimal API, such as `spec.hard` and the basic status fields, and add fields like `scopes`, `remaining`, and `observedDescendants` later when they are actually needed?

docs/quota.md
mhkarimi1383: This is managed by the RBAC permissions

An example will be

We have three tenants and each of them has sub-teams, So each team could have their own quota too...

docs/quota.md
mhkarimi1383: That's okay, but we will need those later on...

docs/quota.md
mhkarimi1383: Native Resource Qouta does not have understanding of the sub namespaces, maybe we can reuse their implementation by importing k8s code here

docs/quota.md
mhkarimi1383: It is close to the native one, so users won't get confused

docs/quota.md
mhkarimi1383: We can just calculate it per request, So if deployment get more replicas is will be okay

docs/quota.md
mhkarimi1383: Also we will need Prometheus Metrics too

docs/quota.md
gokken-roko: I think we should consider the following case:

```
Namespace c  (hard: cpu=3)            ← subtreeUsed: 2, remaining: 1
├── Namespace d  (hard: cpu=2)        ← own used:0, can use only1, not 2
└── Namespace e  (no limit)           ← own used:2, can use only 1
```

- `Namespace e` needs "quota object" to follow the limit of the root namespace.
    - How are these quota objects created?
- It is hard to keep the tree consistent.

ystkfujii: > I think we should consider the following case:

If we allow quotas on any namespace, we will need a separate discussion about how to keep the hierarchy consistent.

For this thread, however, I would first like to clarify whether we actually need to support quotas on arbitrary namespaces.

> We have three tenants and each of them has sub-teams, So each team could have their own quota too...

I understand this use case. 
However, I think it would be difficult to manage this only with Kubernetes RBAC.

If tenant users or sub-tenant users can freely increase or remove their quotas, those quotas would not work as effective restrictions. 
I think platform administrators should manage the quotas, and tenant users should not be allowed to modify them.

If the quota does not work as a restriction, I am not sure what purpose it serves.
For example, if the goal is to limit a single namespace, the standard ResourceQuota should be enough. 
If the goal is only to see the total usage of child namespaces, metrics would be a better fit than quotas.

If tenant users need to allocate resources to sub-tenant users, we may need a mechanism other than Kubernetes RBAC.

---

However, I think quotas managed by platform administrators for each root namespace are useful.

Could we limit the initial implementation to quotas on root namespaces?
If we later find a concrete need for quotas on intermediate namespaces, we can discuss it after the root namespace quota is implemented.

docs/quota.md
ystkfujii: > Native Resource Qouta does not have understanding of the sub namespaces,

That is correct. 
However, I think we can still reuse part of the native ResourceQuota functionality.

The feature we want to implement can be divided into two steps:

- Step1: Calculate resource usage in each namespace.
- Step2: Aggregate local namespace usage across the subtree for each quota-bearing ancestor.

Using native ResourceQuota for step 1 would likely reduce the amount of code we need to implement.

docs/quota.md
ystkfujii: Hmm, I’m not sure this behaves the same way as native ResourceQuota. 🤔 

Native ResourceQuota also has an eventual-consistency race, but it updates `ResourceQuota.status.used` during admission before allowing the request.

- https://github.com/kubernetes/kubernetes/issues/123434#issuecomment-1964920277

With this proposal, two concurrent requests could read the same subtreeUsed and both be allowed before the reconciler updates it.

To behave more like native ResourceQuota, I think we need a way to reserve or atomically update the usage before allowing the request.

docs/quota.md
gokken-roko: > For this thread, however, I would first like to clarify whether we actually need to support quotas on arbitrary namespaces.

I also think this is an important point to decide first.
I personally think limiting the scope to root namespaces is useful.

docs/quota.md
mhkarimi1383: I'm totally agree with that

docs/quota.md
mhkarimi1383: @ystkfujii You're right, and the "identical to native ResourceQuota race" framing in §7 was inaccurate. Native `ResourceQuota` is **not** best-effort on the *increase* path the way my original design was: its in-process admission plugin synchronously increments `ResourceQuota.status.used` via optimistic concurrency (`resourceVersion` CAS) with a retry-on-conflict loop *before* letting the request through admission controllers "may mutate related resources as part of request processing to do things like increment quota usage."【turn0search5】 My design read a cached `subtreeUsed`, projected against it, and deferred the status write to the async reconciler, so two concurrent creates could both read the same stale value and both be admitted. That is a strictly larger race than native RQ's, and I'll fix it.

## Why native RQ is tighter than I claimed

The resourcequota admission plugin treats `status.used` as a reservation ledger, not just an observation:

1. Compute the resource delta of the incoming object.
2. Load the namespace's `ResourceQuota` (cached).
3. Check `used + delta ≤ hard` per key; reject on breach.
4. `PUT /status` carrying the `resourceVersion` from step 2, an optimistic-concurrency increment of `status.used`. Kubernetes uses optimistic concurrency control via `resourceVersion` for exactly this kind of concurrent update.【turn1search5】
5. On `409 Conflict`, reload and retry (bounded). The retry mechanism is confirmed by kubernetes/kubernetes#38447, where batch pod creation hit 409s because "many goroutines update resource quota status concurrently" and the admission controller already retries.【turn1search1】
6. Only after the increment succeeds is the request admitted.

The eventual-consistency race native RQ *does* have is on the **decrease** side: `status.used` lags real cluster state (a pod being terminated is still counted until the quota controller reaps it), so a rolling update's surge pod can be rejected against stale `used`, the kubernetes/kubernetes#89209 / #123434 window you referenced.【turn2search2】 That is *not* a race on concurrent increases; those are serialized by the CAS in step 4. My original design had no equivalent of step 4, so it was racy on increases too.

## Proposed fix: admission-time reservation on each ancestor HRQ

Mirror native RQ's pattern, generalised to the ancestor chain. The webhook becomes a *reserving* webhook (side effect on `status`), not a pure validator.

```text
// On create/update of a quota-scoped object in namespace N
func Admit(obj, N):
    delta := evaluator.Usage(obj)                 // corev1.ResourceList
    if delta is empty: return allow

    chain := ancestorsWithHRQ(N)                  // [N, parent(N), ..., root], HRQ-bearing only
    reserved := []                                // for rollback

    for hrq in chain:                             // root-first
        for attempt in 1..MAX_RETRY (e.g. 5):
            cur := get(hrq)                       // fresh read w/ resourceVersion
            projected := add(cur.status.subtreeUsed, delta)
            if any key: projected > cur.spec.hard:
                rollback(reserved)                // undo increments already made
                return reject(403, hrq, key, projected, hard)
            err := putStatus(hrq,
                             status.subtreeUsed = projected,
                             resourceVersion      = cur.resourceVersion)
            if err is 409 Conflict: continue      // reload & retry
            if err != nil:
                rollback(reserved)
                return reject(500, "quota reservation failed", hrq)
            reserved.append((hrq, delta))
            break
        else:                                     // retries exhausted
            rollback(reserved)
            return reject(429, "quota contention, retry", hrq)

    return allow
```

Key properties:

- Each ancestor HRQ is incremented with its own CAS, so two admissions sharing an ancestor serialize on *that* object's `resourceVersion`, the second admission's CAS fails, it reloads, sees the first's increment, and re-checks. This is native RQ's serialization generalised per-ancestor.
- Rollback on partial failure keeps the ledger from leaking; if rollback itself races, the periodic full-resync reconciler corrects drift, exactly as the core `resourcequota` controller periodically overwrites `status.used` from real observed usage.【turn0search8】
- `failurePolicy=Fail` remains; retry exhaustion rejects (fail-closed).

## The reconciler's role (now matching native RQ)

- **Decreases and drift:** `HierarchicalResourceQuotaReconciler` stays the source of truth for `status.used` (real observed usage) and recomputes `status.subtreeUsed = used + Σchildren.subtreeUsed` on resync. This corrects speculative-increment drift and accounts for deletes, identical to the core `resourcequota` controller periodically recomputing `status.used`.
- **Transient dip:** because the reconciler overwrites `subtreeUsed` from `used + children`, a just-admitted pod whose usage hasn't yet hit the usage informer could briefly make `subtreeUsed` dip below the reserved value, reopening a tiny window. Native RQ has the same dip between the admission increment and the controller recompute; it's bounded and the next reconcile closes it. If we want to eliminate it entirely, a `status.reserved` field that the webhook adds to and the reconciler drains only as `used` catches up would do it, I'd like your call on whether that extra field is worth the precision.

## Residual race after the fix

| Path | Native RQ | HRQ (fixed) |
|------|-----------|-------------|
| Concurrent *increases* in the same namespace | Serialized by CAS on the single RQ | Serialized by CAS on the local HRQ |
| Concurrent *increases* in sibling sub-namespaces sharing an ancestor budget | N/A (single namespace) | Serialized by CAS on the shared ancestor HRQ |
| *Decrease* lag (terminating pods, async deletes) | Yes, #89209 / #123434【turn2search2】 | Yes, same, reconciler-driven |
| Multi-ancestor atomicity (N increments, not a txn) | N/A (one object) | Not atomic across ancestors, but fail-closed: a partial reservation that can't extend to a tighter ancestor rolls back; worst case is a transient false-reject, never over-admission |

After the fix the residual race is precisely the decrease-lag race native RQ already has, not the increase race my original design introduced.

## Spec/controller deltas

- Webhook reclassified from pure `Validating` to `sideEffects: NoneOnDryRun` (it writes `status`); `dryRun` requests skip the reservation.
- `status.subtreeUsed` becomes writable by both the webhook (speculative increments, SSA field manager `hrq-webhook`) and the reconciler (authoritative recompute, SSA field manager `hrq-reconciler`); server-side-apply field ownership keeps the two writers from stomping each other, with the reconciler authoritative on resync.
- New `conditions: QuotaContention` set when the retry budget is exhausted.

## Alternative considered and rejected

A single cluster-scoped `HierarchicalResourceQuotaTree` object holding all subtree usages as a map, so admission is a single CAS (true cross-ancestor atomicity). Rejected as default because every admission in the entire tree would serialize on one object — a throughput cliff for large tenants. Per-ancestor CAS distributes contention along the tree shape and matches native RQ's per-namespace granularity. The single-object form can be offered later as `enforcementMode: tree-locked` for small trees that want strict atomicity.

I'll fold the pseudocode and the residual-race table into §7 and add the `reserved`-field question to §11. Does this reservation model match what you had in mind, or would you prefer the single-tree-object variant for stricter atomicity?

docs/quota.md
mhkarimi1383: @ystkfujii @gokken-roko Agreed, scoping the initial implementation to **root-namespace-only HRQ** is the right call. The original #237 request was per-root, intermediate quotas introduce a delegation problem that RBAC alone doesn't solve, and HNC's own HRQ design treats the namespace holding the HRQ as the authoritative budget for its whole subtree, which is the same shape we'd ship. 【turn3fetch0】 Below is the revised scope, the answer to @gokken-roko's consistency scenario, and the Phase 2 sketch.

## Decision: Phase 1 = root-only

- An `HierarchicalResourceQuota` may only be created in a namespace labeled `accurate.cybozu.com/type=root`. A validating webhook on HRQ create rejects placement elsewhere (and rejects a second HRQ in the same tree).
- Exactly one authoritative HRQ per tree. The admission webhook, on any quota-scoped create/update in any namespace in the tree, walks up `accurate.cybozu.com/parent` to the root and enforces against that single HRQ's `subtreeUsed + delta ≤ hard`.
- The admission-time reservation fix from the previous round still applies, but simplifies to a **single CAS per tree** (on the root HRQ's `status.subtreeUsed`), since there are no intermediate HRQs to chain.
- No changes to the `SubNamespace` CRD. Tenant users keep creating/deleting sub-namespaces freely; all usage rolls up to the one root budget.

## Why this resolves @gokken-roko's scenario

The c/d/e tree under root-only no longer has the inconsistency:

```
Namespace c  (HRQ hard: cpu=3)   ← the only quota object in the tree
├── Namespace d  (no HRQ)
└── Namespace e  (no HRQ)
```

| Admission request | Webhook behavior |
|---|---|
| Pod in `e` requesting cpu=1 (e.used=2) | Walk to root `c`; project `subtreeUsed = 2 + 1 = 3 ≤ 3` → admit |
| Pod in `e` requesting cpu=2 | Project `4 > 3` → reject against `c`'s HRQ |
| Pod in `d` requesting cpu=2 (d.used=0) | Project `0 + 2 = 2 ≤ 3` → admit (d has no HRQ, so no per-namespace cap; only the tree cap applies) |

@gokken-roko's two concerns dissolve:

1. **"Namespace e needs a quota object to follow the root's limit."** It doesn't. Enforcement is centralized at the root HRQ and applied transitively by the webhook. No object needs to exist in `e`; no propagation of quota objects into children. This is the core simplification of root-only.
2. **"It is hard to keep the tree consistent."** There's nothing to keep consistent: one HRQ, one `subtreeUsed` counter, one CAS target. The reparenting/cut cases only require re-enqueuing the root HRQ so it re-sums its (changed) descendant set, no per-child `effectiveHard` to recompute, no sibling-dependent capacity thrash.

The inconsistency in the original proposal arose specifically from `d` having `hard: cpu=2` while only 1 was effectively usable — that's the dynamic-`effectiveHard` problem, and it only exists when intermediate HRQs are allowed. Dropping intermediate HRQs removes the problem entirely.

## Why RBAC is sufficient for root-only but not for delegation

For root-only, RBAC gives a clean ownership boundary, and it lines up with how native `ResourceQuota` is already protected:

- The built-in `admin` ClusterRole explicitly **excludes write access to resource quota** — tenant users granted `admin` via the propagated RoleBinding cannot modify quotas. 【turn0search2】
- We mirror this for HRQ: the HRQ admin `ClusterRole` is **not** labeled `rbac.authorization.k8s.io/aggregate-to-admin`, so it does not aggregate into the tenant `admin` role. Only platform admins get a separate RoleBinding granting HRQ CRUD. The propagated tenant RoleBinding (`accurate.cybozu.com/propagate=update`, giving tenant `admin` on all sub-namespaces) 【turn0search0】 therefore never conveys HRQ write.
- Net: platform admin owns the one root HRQ; tenant users own their sub-namespaces and workloads; neither can touch the other's domain. Standard RBAC, no extra machinery.

For **intermediate quotas**, this breaks down. To let a sub-team set its own sub-budget you must grant that sub-tenant HRQ write in their sub-namespace — at which point nothing in RBAC stops them from setting `hard: cpu=1000` or deleting the HRQ entirely. @ystkfujii's point is exactly this: a quota the subject can remove or raise is not a restriction, it's a suggestion. Making delegation work needs something beyond RBAC — a bound that ties a child HRQ's `hard` to a slice the parent granted.

## Phase 2 sketch (deferred, separate design)

Intermediate-quota delegation is a real feature, but it needs its own design. The most credible prior art is HNC's, which solves the ownership gap with a **structural** pattern rather than RBAC alone: 【turn3fetch0】

- A dedicated **policy namespace** sits above the tenant root and holds the platform-admin-owned HRQ (`company-a-policy → company-a`). The tenant cannot modify the parent policy but can create sub-HRQs in their own policy namespaces.
- The rule **"child quotas cannot override more restrictive quotas from ancestor namespaces; the most restrictive quota always applies"** gives the bound that RBAC can't — a child HRQ is effectively clamped by every ancestor's HRQ. 【turn3fetch0】
- HNC implements this by writing internal `ResourceQuota` objects (`hrq.hnc.x-k8s.io`) into each affected namespace, i.e., the dynamic-`hard`-rewrite approach (Alternative B in our §10). That's the part we'd want to redesign (it has the `hard`-thrash and over-admission-on-shrink problems we already flagged), likely via the admission-reservation model generalized to the ancestor chain rather than per-namespace RQ rewriting.

Concretely, Phase 2 would need to specify at minimum:

1. **Ownership/bound model** — who may create an intermediate HRQ, and the invariant `child.spec.hard ≤ parent.allocated(child)` enforced by a validating admission policy or controller clamp. 【turn0search13】
2. **Most-restrictive-wins** semantics on admission (min over ancestor HRQs), with `status.effectiveHard` for observability.
3. **Tree-mutation rules** — what happens to an intermediate HRQ on `sub move` under a tighter parent, on `sub cut` to root, on cascading delete.
4. **Borrowing** — whether a child may temporarily exceed its `hard` against unused parent headroom.

These are orthogonal to the root-only enforcement engine and can be layered on later without reworking Phase 1. The Phase 1 webhook's ancestor-walk already generalizes to "enforce against the min over ancestor HRQs"; the delta for Phase 2 is the ownership/bound model, not the enforcement path.

## Design-doc deltas for Phase 1

| Section | Change |
|---|---|
| §4.1 CRD | Add validating rule: HRQ create rejected unless `metadata.namespace` has `accurate.cybozu.com/type=root`. Drop the multi-HRQ-per-tree allowance. |
| §4.2 Semantics | Remove "effective ceiling = min over ancestors"; replace with "the root HRQ is the sole ceiling for the entire subtree." Keep `status.subtreeUsed` roll-up. |
| §4.3 Controller | `tree.Ancestors` simplifies to `tree.Root(ns)`; reservation CAS targets the single root HRQ. |
| §4.4 Enforcement | Webhook walks to root, enforces against root HRQ only. No intermediate HRQs to consider. |
| §4.6 Walk-through | Rewrite the example: `a` root-only with HRQ; `b`/deeper sub-namespaces have no HRQ; `c` root-only with HRQ; `d`/`e` no HRQ. The "b capped at 5 under a capped at 10" row moves to Phase 2. |
| §9 Rollout | Feature gate `HierarchicalResourceQuota` ships root-only. Add `HierarchicalResourceQuotaDelegation` as a separate, future feature gate. |
| §10 Alternatives | Add HNC's policy-namespace + most-restrictive-wins as the reference approach for Phase 2, noting its internal-RQ-rewrite implementation and why we'd reservation-based instead. 【turn3fetch0】 |
| §11 Open questions | Replace the intermediate-quota items with: "Phase 2: ownership model for delegated HRQs; most-restrictive-wins vs. reservation-based enforcement; borrowing." |

The user-facing YAML in §5 keeps the root HRQs (`a`, `c`) verbatim and drops the per-sub-namespace HRQs (`b`, `d`); the `SubNamespace` manifests are unchanged.

@gokken-roko, does the root-only enforcement model (single HRQ per tree, webhook walks to root, no child quota objects) address the consistency concern you raised? And @ystkfujii, is root-namespace-only with a clearly separated Phase 2 delegation track an acceptable scope for the initial PR?

mhkarimi1383: We can Also Add admission hooks for native resourcequota too,

Reject it on root namespace if we already have Our own resourcequota
Reject it on sub-namespaces if the limit is higher than root namespace limit

gokken-roko: Thank you. Yes, I think limiting the scope to root namespace resolves my concerns.
I personally think this change affects the overall design.  For this topic, I would like to discuss the details after the design doc is updated.

mhkarimi1383: Okay, No problem

mhkarimi1383:
Hi, Let me know when your done with reviews, So I will update the document again

ystkfujii:
@mhkarimi1383 
Sorry for the delayed response.

After considering this proposal, we have decided not to move forward with it at this time. We think that this functionality is beyond the current scope of Accurate and would add too much complexity to the project.

In Accurate, sub-namespaces are represented by labels on Namespace resources. Supporting ResourceQuota across sub-namespaces would therefore require grouping namespaces by their labels and aggregating their resource usage. This kind of mechanism is not specific to Accurate and could be useful in other use cases as well. For this reason, we think it would be better developed as a standalone open-source project or proposed as an enhancement to Kubernetes ResourceQuota itself.

Thank you very much for taking the time to consider ResourceQuota support for Accurate and for sharing this proposal.

PR #239, opened 2026-09-08
kubebuilder v4 layout (partially)

ystkfujii:
follow up : https://github.com/cybozu-go/accurate/pull/236

Accurate already differs from the standard Kubebuilder layout, so we do not need to follow it completely.
However, I think adopting some parts of the Kubebuilder layout can improve the current package structure.

The main goals of this PR are:

- Use internal packages to separate internal implementation.
  - This pattern is also described in the official Go documentation.
  - ref: https://go.dev/doc/modules/layout#package-or-command-with-supporting-packages
- Group metadata-related code into the v2 package.
  - plz see: https://github.com/cybozu-go/accurate/pull/236#issuecomment-4947563202

There may be different opinions on the other parts of the package structure.

For those parts, I would like to use the Kubebuilder layout as a common baseline so that we can more easily reach agreement.


012xx (approved):
I ran tests in a local environment and confirmed that they were working properly.
LGTM

012xx:
@ystkfujii 
It's difficult to decide whether to apply this partially or fully, but I thought it would be fine as is.

Could you once again add to the PR description the criteria used for this decision and the reasons why it ended up in this partial-adoption state, so that other people can understand it later?

ystkfujii:
@012xx updated the description

PR #241, opened 2026-09-15
Introduce Takumi Guard

takahiro-yamada:
Setup Takumi Guard for golang to mitigate the risk of software supply chain attacks.


.github/actions/setup-go/action.yaml
ystkfujii, line 1: `setup-go` sounds like an action that only sets up Go, but this action also configures Takumi Guard.

How about using a more explicit name, such as `setup-go-with-takumi-guard`? 
It may be a bit long, so I’m also fine with another name that makes the purpose clear.

Alternatively, I think it would be fine not to wrap these steps in a composite action.

.github/workflows/ci.yaml
ystkfujii, line 1: I don't think we should enable Takumi Guard for CI triggered by pull requests.

For pull requests from external contributors, the workflow will likely fail because OIDC permissions are restricted for forked repositories.

.github/actions/setup-go-with-takumi-guard/action.yaml
takahiro-yamada: I agree, fixed: [e9e327a](https://github.com/cybozu-go/accurate/pull/241/commits/e9e327a5a6707b71f14970dd898b93db09a7131e)

.github/workflows/ci.yaml
takahiro-yamada: fixed: [e9e327a](https://github.com/cybozu-go/accurate/pull/241/commits/e9e327a5a6707b71f14970dd898b93db09a7131e)

.github/workflows/ci.yaml
```diff
@@ -4,32 +4,28 @@ on:
...
     - 'main'
-permissions:
-  contents: read
 jobs:
   build:
     name: Build binaries
     runs-on: ubuntu-24.04
+    permissions:
+      contents: read
+      id-token: write
```
ystkfujii: `id-token` might not be necessary.

```suggestion
```

.github/workflows/ci.yaml
```diff
@@ -43,15 +39,15 @@ jobs:
         - v1.35.5@sha256:ce977ae6d65918d0b58a5f8b5e940429c2ce42fa3a5619ec2bbc60b949c0ac95 # renovate: kindest/node
         - v1.34.8@sha256:02722c2dedddcfc00febf5d27fbeb9b7b2c14294c82109ff4a85d89ac9ba3256 # renovate: kindest/node
     runs-on: ubuntu-24.04
+    permissions:
+      contents: read
+      id-token: write
```
ystkfujii: ditto

https://github.com/cybozu-go/accurate/pull/241/changes#r4043041424

.github/workflows/release.yaml
```diff
@@ -31,15 +31,13 @@ jobs:
     runs-on: ubuntu-24.04
     permissions:
       contents: write
+      id-token: write
```
ystkfujii: ditto

https://github.com/cybozu-go/accurate/pull/241/changes#r4043041424

.github/workflows/ci.yaml
takahiro-yamada: Thanks, fixed: [c1c4e66](https://github.com/cybozu-go/accurate/pull/241/commits/c1c4e66d1a844e0a5f99b3aca996e85800a761ba)

.github/workflows/ci.yaml
takahiro-yamada: [c1c4e66](https://github.com/cybozu-go/accurate/pull/241/commits/c1c4e66d1a844e0a5f99b3aca996e85800a761ba)

.github/workflows/release.yaml
takahiro-yamada: [c1c4e66](https://github.com/cybozu-go/accurate/pull/241/commits/c1c4e66d1a844e0a5f99b3aca996e85800a761ba)

PR #245, opened 2026-09-17
Update chart version to 0.8.0

ystkfujii:
Ref: https://github.com/cybozu-go/accurate/issues/173#issuecomment-5711120396

Update the Helm chart version from 0.7 to 0.8 in preparation for the v2.0.0 release.

With the current release workflow, releasing v2.0.0 should result in:

```
version: 0.8.0
appVersion: 2.0.0
```

If the result is different, I plan to create a follow-up PR to fix it.


takahiro-yamada (approved):
LGTM

erikgb:
@ystkfujii, why are we not aligning versions? Meaning the next will be 2.0.0? I've always found the mismatch of versions confusing. Did we ever release just the chart?

ystkfujii:
@erikgb 

Sorry for the late replay. 🙇 

The chart and application versions have different meanings, so I don't think they necessarily need to match. 
However, since Accurate currently releases them together, I agree that aligning the versions, like cert-manager does, would be clearer.

i'll change the chart version to match the app version.
- https://github.com/cybozu-go/accurate/pull/256

